Issue #23, opened 2014-12-16
Use cached_property decorator 


iskandr:
Currently we're checking cached properties either by setting the data to None in the initializer:
```
_length = None
```
and then checking
```
if _length is None:
   ...compute...
   self._length = result
 ```

Or, uglier, checking using `hasattr`:
```
if hasattr(self, "_length"):
   ...compute...
   self._length = result
```

The former sucks because pieces of the program can get out of sync. The latter is ugly. 

Better would be to use something like:

```
@cached_property
def length(self):
   return some_computation()
```

PR #26, opened 2014-12-23
changed @property (with ad-hoc hasattr caching) to instead use memoized-property library

iskandr:
Fixes https://github.com/hammerlab/pyensembl/issues/23

Also:
* check validity of gene_biotype on Gene object
* sped up `__contains__` and sequence access on `ReferenceTranscripts` by by-passing `pyfaidx.Fasta.__contains__` (which copies a list of keys every time) 

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/26)

### DIFF
--- a/pyensembl/gene.py
+++ b/pyensembl/gene.py
@@ -68,11 +68,10 @@ class Gene(Locus):
         # We're doing a SQL query for each transcript ID to fetch
         # its particular information, might be more efficient if we
         # just get all the columns here, but how do we keep that modular?
-        transcripts = [
+        return [
             Transcript(result[0], self.db)
             for result in results
         ]
-        self._transcripts = transcripts
 
     @memoized_property
     def exons(self):
--- a/pyensembl/gene.py
+++ b/pyensembl/gene.py
@@ -1,5 +1,8 @@
+from biotypes import all_valid_biotypes
 from locus import Locus
 from transcript import Transcript
+
+from memoized_property import memoized_property
 
 class Gene(Locus):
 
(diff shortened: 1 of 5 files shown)


pyensembl/gene.py
```diff
@@ -43,40 +50,36 @@ def __str__(self):
...
-            self._transcripts = transcripts
-        return self._transcripts
+        # We're doing a SQL query for each transcript ID to fetch
+        # its particular information, might be more efficient if we
+        # just get all the columns here, but how do we keep that modular?
+        transcripts = [
+            Transcript(result[0], self.db)
+            for result in results
+        ]
+        self._transcripts = transcripts
```
ihodes: I think you may be forgetting the `return self._transcripts`?

iskandr: Good catch, this code actually works for arcane reasons (memoized_property will run this function to set self._transcripts and then return the value found there). Obviously need to change it to be sane.

pyensembl/transcript.py
```diff
@@ -69,46 +69,40 @@ def __len__(self):
...
+        exons = [None] * len(results)
 
-                exon = Exon(exon_id, self.db)
-                exon_number = int(exon_number)
-                assert exon_number >= 1, "Invalid exon number: %s" % exon_number
-                assert exon_number <= len(exons), \
-                    "Invalid exon number: %s (max expected = %d)" % (
-                        exon_number, len(exons))
+        for entry in results:
+            exon_number, exon_id = entry
```
ihodes: Just FWIW you could do this destructuring in the loop, `for exon_number, exon_id in results:`

ihodes:
LGTM % that one odd return statement

PR #34, opened 2015-02-12
Make downloads explicit

### LINKED ISSUE #31: Make data downloads explicit
Instead of transparently downloading data the first time it's needed (which introduces unexpected delays), instead require the user to download a data source explicitly. 

```
ensembl = EnsemblRelease()
ensembl.download_annotations()
ensembl.download_transcript_sequences()
```

When a methods requires a data source which isn't yet available, raise an exception and tell the user to call the required download method.

tavinathanson:
Make downloads explicit (unless `auto_download` is `True`). An Exception is raised if the necessary data isn't present, and users can either use the shell or python to run the downloads. Fixes #31 

Relies on https://github.com/hammerlab/datacache/pull/7

A few questions that come to mind:

- Did I hook into the right places?
- Do you think the `varcode` should also have an `auto_download` flag to send along to `pyensembl`?

TODO:

- Tests
- README update

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/34)

### DIFF
--- a/pyensembl/database.py
+++ b/pyensembl/database.py
@@ -16,8 +16,9 @@ class Database(object):
     writing SQL queries directly.
     """
 
-    def __init__(self, gtf):
+    def __init__(self, gtf, auto_download=False):
         self.gtf = gtf
+        self.auto_download = auto_download
         self._connection = None
 
     def __eq__(self, other):
@@ -70,7 +71,7 @@ class Database(object):
         indices.append(['feature'])
         return indices
 
-    def _create_database(self):
+    def _create_database(self, force=False):
         print("Creating database: %s" % self.local_db_path())
         filename = self.local_db_filename()
         df = self.gtf.dataframe()
@@ -83,33 +84,42 @@ class Database(object):
             table_name="ensembl",
             df=df,
             subdir=CACHE_SUBDIR,
-            overwrite=False,
+            overwrite=force,
             indices=indices)
         return db
 
-    def _connect_or_create_database(self):
+    def _connect_if_exists(self):
         """
-        If database already exists, open a connection.
-        Otherwise, create it.
+        Return the connection if the DB exists, and otherwise return 
+        None. As a side effect, stores the database connection in 
+        self._connection.
         """
-        db_path = self.local_db_path()
-        if exists(db_path):
-            db = sqlite3.connect(db_path)
-            # maybe file got created but not filled
-            if datacache.db.db_table_exists(db, 'ensembl'):
-                return db
-        return self._create_database()
+        if self._connection is None:
+            db_path = self.local_db_path()
+            if exists(db_path):
+                connection = sqlite3.connect(db_path)
+                # maybe file got created but not filled
+                if datacache.db.db_table_exists(connection, 'ensembl'):
+                    self._connection = connection
+        return self._connection
 
     @property
     def connection(self):
         """
         Return the sqlite3 database for this Ensembl release
-        (download and/or construct it necessary).
-        As a side effect, stores the database connection in self._db
+        (download and/or construct it if necessary, if auto_download
+        is on). As a side effect, stores the database connection in 
+        self._connection.
         """
-        if self._connection is None:
-            self._connection = self._connect_or_create_database()
-        return self._connection
+        if self._connect_if_exists():
+            return self._connection
+        if self.auto_download:
+            return self._create_database()
+        raise ValueError("Ensembl annotations data is not currently "
+                         "installed for release %s. Run "
+                         "\"pyensembl install %s\" or call into "
+                         "EnsemblRelease(%s).install()" %
+                         ((self.gtf.release,) * 3))
 
     def columns(self):
         sql = "PRAGMA table_info(ensembl);"
@@ -386,3 +396,18 @@ class Database(object):
             raise ValueError("Too many loci for %s with %s = %s: %s" % (
                 feature, filter_column, filter_value, loci))
         return loci[0]
+
+    def create(self, force=False):
+        """
+        Create the local database (including indexing) if it's not
+        already set up. If `force` is True, always re-create
+        the database from scratch.
+
+        Returns True if the database was re-created.
+
+        Raises an error if the necessary data is not yet downloaded.
+        """
+        if not force and self._connect_if_exists():
+            return False
+        self._create_database(force=force)
+        return True
--- a/pyensembl/ensembl_release.py
+++ b/pyensembl/ensembl_release.py
@@ -9,7 +9,6 @@ import logging
 from os.path import join
 from os import remove
 
-from .common import CACHE_SUBDIR
 from .compute_cache import cached_object
 from .database import Database
 from .exon import Exon
@@ -20,10 +19,6 @@ from .release_info import check_release_number, MAX_ENSEMBL_RELEASE
 from .transcript import Transcript
 from .url_templates import ENSEMBL_FTP_SERVER
 
-import datacache
-import numpy as np
-import pandas as pd
-
 
 class EnsemblRelease(object):
     """
@@ -32,15 +27,22 @@ class EnsemblRelease(object):
     variety of helper methods for accessing this data.
     """
 
-    def __init__(self, release=MAX_ENSEMBL_RELEASE, server=ENSEMBL_FTP_SERVER):
-        self.cache = datacache.Cache(CACHE_SUBDIR)
+    def __init__(self,
+                 release=MAX_ENSEMBL_RELEASE,
+                 server=ENSEMBL_FTP_SERVER,
+                 auto_download=False):
         self.release = check_release_number(release)
         self.species = "homo_sapiens"
         self.server = server
-        self.gtf = GTF(self.release, self.species, server)
-        self.db = Database(gtf = self.gtf)
+        self.auto_download = auto_download
+        self.gtf = GTF(self.release, self.species, server,
+                       auto_download=auto_download)
+        self.db = Database(gtf=self.gtf, auto_download=auto_download)
         self.reference = ReferenceTranscripts(
-            self.release, self.species, server)
+            ensembl_release=self.release,
+            species=self.species,
+            server=server,
+            auto_download=auto_download)
 
         self.logger = logging.getLogger()
         self.logger.setLevel(logging.INFO)
@@ -68,7 +70,6 @@ class EnsemblRelease(object):
         self.gtf.clear_cache()
         self.reference.clear_cache()
         self._delete_cached_files()
-
 
     def all_feature_values(
             self,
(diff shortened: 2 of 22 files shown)


pyensembl/ensembl_release.py
```diff
@@ -120,6 +123,33 @@ def run_query():
...
+    def download_annotations(self):
+        if self.db.connection_if_exists():
+            print ("Annotation data for release %s is already "
+                   "downloaded and installed" % self.release)
+            return
+        self.db.create_database()
+        print ("Annotation data for release %s has been downloaded "
+               "and installed" % self.release)
+
+    def download_transcripts(self):
```
tavinathanson: I guess `transcripts` is a bit of a misnomer. `transcript_sequences`, as you originally suggested, might be better.

pyensembl/database.py
```diff
@@ -81,25 +82,42 @@ def _create_database(self):
...
         if exists(db_path):
-            db = sqlite3.connect(db_path)
+            connection = sqlite3.connect(db_path)
             # maybe file got created but not filled
-            if datacache.db.db_table_exists(db, 'ensembl'):
-                return db
-        return self._create_database()
+            if datacache.db.db_table_exists(connection, 'ensembl'):
+                return connection
+        return None
```
tavinathanson: Unnecessary line

pyensembl/ensembl_release.py
```diff
@@ -120,6 +123,33 @@ def run_query():
             return results
         return cached_object(pickle_path, compute_fn=run_query)
 
+    def download_all(self):
+        self.download_annotations()
+        self.download_transcripts()
+
+    def download_annotations(self):
+        if self.db.connection_if_exists():
```
tavinathanson: Oops: it looks like I'm not closing the connection or saving it anywhere.

pyensembl/database.py
```diff
@@ -64,7 +65,7 @@ def _database_indices(self, available_columns):
         indices.append(['feature'])
         return indices
 
-    def _create_database(self):
+    def create_database(self):
         print "Creating database: %s" % self.local_db_path()
```
iskandr: In alliance with Tim's attempt to Python3 all the things, can we wrap this print in parentheses?

iskandr: In alliance with Tim's attempt to Python3 all the things, can we wrap this print in parentheses?

pyensembl/database.py
```diff
@@ -81,25 +82,42 @@ def _create_database(self):
...
+        connection = self.connection_if_exists()
+        if connection:
+            return connection
+        if self.auto_download:
+            return self.create_database()
+        raise ValueError("Ensembl annotations data is not currently "
+                         "downloaded for release %s. Run "
+                         "\"pyensembl update %s\" or call into "
+                         "EnsemblRelease(%s).download_annotations()" %
+                         tuple([self.gtf.release] * 3))
```
iskandr: Small nit: You can do `(self.gtf.release,) * 3`.

pyensembl/database.py
```diff
@@ -81,25 +82,42 @@ def _create_database(self):
...
+
+    def _connect_or_create_database(self):
+        """
+        If database already exists, open a connection.
+        Otherwise, create it if auto download is enabled.
+        """
+        connection = self.connection_if_exists()
+        if connection:
+            return connection
+        if self.auto_download:
```
iskandr: Is there any way to force a re-download?

pyensembl/database.py
```diff
@@ -81,25 +82,42 @@ def _create_database(self):
             indices=indices)
         return db
 
-    def _connect_or_create_database(self):
+    def connection_if_exists(self):
```
iskandr: It seems a little odd that `_connect_or_create_database` is still "private" but this method is "public" (doesn't start with an underscore). Seems like they should be equally exposed to the user (in either direction).

pyensembl/ensembl_release.py
```diff
@@ -14,28 +14,32 @@
...
+                 release=MAX_ENSEMBL_RELEASE,
+                 server=ENSEMBL_FTP_SERVER,
+                 auto_download=False):
         self.cache = datacache.Cache(CACHE_SUBDIR)
         self.release = check_release_number(release)
         self.species = "homo_sapiens"
         self.server = server
         self.gtf = GTF(self.release, self.species, server)
-        self.db = Database(gtf = self.gtf)
+        self.auto_download = auto_download
```
iskandr: When does EnsemblRelease need to know about `auto_download`?

pyensembl/ensembl_release.py
```diff
@@ -120,6 +123,33 @@ def run_query():
             return results
         return cached_object(pickle_path, compute_fn=run_query)
 
+    def download_all(self):
+        self.download_annotations()
+        self.download_transcripts()
+
+    def download_annotations(self):
```
iskandr: Can we add `force_download` options to these?

pyensembl/ensembl_release.py
iskandr: Yeah, I think there will be lots of other methods where the `_transcripts` suffix suggests we're returning a list of `Transcript` objects.

pyensembl/ensembl_release.py
```diff
@@ -120,6 +123,33 @@ def run_query():
...
+            print ("Annotation data for release %s is already "
+                   "downloaded and installed" % self.release)
+            return
+        self.db.create_database()
+        print ("Annotation data for release %s has been downloaded "
+               "and installed" % self.release)
+
+    def download_transcripts(self):
+        cache = self.reference.cache
+        if cache.exists(self.reference.url,
```
iskandr: Shouldn't this logic be on the ReferenceTranscripts object? All this `self.reference.foo` seems like a code smell (or a hint that this method should move elsewhere).

pyensembl/shell.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python
```
iskandr: How does the user know where to find this script?

setup.py
```diff
@@ -31,12 +31,17 @@
...
         name='pyensembl',
-        version="0.5.3",
+        version="0.5.4",
         description="Python interface to ensembl reference genome metadata",
       	author="Alex Rubinsteyn",
         author_email="alex {dot} rubinsteyn {at} mssm {dot} edu",
       	url="https://github.com/hammerlab/pyensembl",
         license="http://www.apache.org/licenses/LICENSE-2.0.html",
+        entry_points={
+          'console_scripts': [
```
iskandr: Oh, does this somehow magically add `shell.py` to PATH?

pyensembl/database.py
tavinathanson: I made this one "public" because it's used by `EnsemblRelease`; that makes sense to make `_connect_or_create_database` "public".

pyensembl/database.py
tavinathanson: Good idea. What about situations where we want to force a SQLite setup without re-downloading? That's probably overcomplicating things.

pyensembl/database.py
tavinathanson: :+1:

pyensembl/ensembl_release.py
tavinathanson: The flow I imagined was:

```
data = EnsemblRelease(77)
data.do_something
# exception, you need to download
```

vs.

```
data = EnsemblRelease(77, auto_download=True)
data.do_something
# downloads automatically, as before
```

pyensembl/ensembl_release.py
tavinathanson: Sure, I think that makes sense. Per my comment above, I could imagine multiple levels of granularity for forcing (download vs. DB setup), but I think it'd be simplest to fully delete and start from scratch.

pyensembl/ensembl_release.py
tavinathanson: :+1:

pyensembl/ensembl_release.py
tavinathanson: Sure, I think that makes sense!

setup.py
tavinathanson: It sticks an executable `pyensembl` into the Python install dir, and `pyensembl` looks like:

```
#!/Users/tavi/.virtualenvs/pyensembl/bin/python2.7
# EASY-INSTALL-ENTRY-SCRIPT: 'pyensembl==0.5.4','console_scripts','pyensembl'
__requires__ = 'pyensembl==0.5.4'
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.exit(
        load_entry_point('pyensembl==0.5.4', 'console_scripts', 'pyensembl')()
    )
```

pyensembl/shell.py
tavinathanson: See below

pyensembl/database.py
iskandr: I was actually just worrying about that. Changing the database schema is
high on my TODO list, it would be annoying to have to download to
regenerate the database. Any ideas? Maybe purely local operations can still
happen automatically/magically, with a loud warning to the user?

tavinathanson:
@iskandr Ready for review -- I think this addresses what we talked about: install, index, download -- let me know what you think

pyensembl/ensembl_release.py
```diff
@@ -125,6 +126,57 @@ def run_query():
...
+        Explicitely index all data for this release, regardless of
+        whether it is already indexed. Raises an error if data is
+        not downloaded.
+        """
+        self._index(force=True)
+
+    def _index(self, force):
+        if self.db.create(force=force):
+            print("Annotation data for release %s has just been indexed" %
+                  self.release)
```
iskandr: I'm less interested in seeing status update *after* some computationally intensive work than before. I would probably change this to logging.info("...")

pyensembl/database.py
```diff
@@ -83,33 +83,42 @@ def _create_database(self):
...
+        (download and/or construct it if necessary, if auto_download
+        is on). As a side effect, stores the database connection in 
+        self._connection.
         """
-        if self._connection is None:
-            self._connection = self._connect_or_create_database()
-        return self._connection
+        if self._connect_if_exists():
+            return self._connection
+        if self.gtf.auto_download:
```
iskandr: It seems strange that you need to reach into the GTF object to check whether it auto-downloads. Can't you raise the error below inside of GTF if auto_download=False?

pyensembl/ensembl_release.py
tavinathanson: Done

pyensembl/database.py
tavinathanson: Hmm. This error is raised if there's no connection to the DB *and* auto download is off.

There's a similar code path in the GTF object (see `local_gtf_path`), and the reason I have them both is that I wanted to raise an error if (a) there's no cached GTF file and (b) if there's no DB.

(I might have a GTF file without a DB, and don't want to auto-create the DB. But, if I just have the DB code path, it auto-downloads on force `index`.)

For now, I added an `auto_download` in `database.py` so I don't have to reach into GTF. Maybe `auto_download` is confusing, since it covers both downloading and indexing, but `auto_install` seems more confusing.

Let me know if you have more thoughts, though. I might be misunderstanding your suggestion.

tavinathanson: In general, I think these code paths are too fragile and confusing. Totally open to suggestions for fixing that.

iskandr: >I think these code paths are too fragile and confusing.

I agree, I was already uncomfortable with the mental complexity of these paths and now it seems even trickier to follow. However, I'd suggest merging this PR and then thinking about paring back the complexity when we get frustrated with it later.

tavinathanson:
@iskandr I'm not clear from "However, I'd suggest merging this PR": is this ready to merge?

tavinathanson:
@iskandr the merge conflict was just a `setup.py` conflict that a rebase resolved

PR #35, opened 2015-02-13
Support Python 3

timodonnell:
With these changes, the unit tests run for me in both Python 3.4 and Python 2.7.

Changes that affect the code running under Python 2:
 * unicode values are converted to str in Database.py. Everywhere that used to assert that a value was either a unicode or a str now asserts it's a str.
 * Many places that assert a numeric value is an int or a long now require it to be an int. A Python 2 int is a C long, so it's unlikely we're going to actually need Python 2 long values, but if there's a case where this is needed we will have to revisit.
 * Fixed an apparent bug in test_exon_object.py in the test_partial_start_codons method.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/35)


pyensembl/compute_cache.py
```diff
@@ -8,7 +8,7 @@
 2) Cache warm on disk. Parse or unpickle the serialized result into memory.
 3) Cache warm in memory. Return cached object.
 """
-import cPickle
+import pickle
```
iskandr: Wait wait wait, this has big performance implications for Python 2.7.x. 

```
In [1]: d = {'a': range(100), 'b': [str(i) for i in range(100)]}
In [2]: timeit s=pickle.dumps(d, pickle.HIGHEST_PROTOCOL)
1000 loops, best of 3: 523 µs per loop
In [3]: timeit s=cPickle.dumps(d, cPickle.HIGHEST_PROTOCOL)
10000 loops, best of 3: 22.2 µs per loop
```

pyensembl/__init__.py
```diff
@@ -1,5 +1,5 @@
-from ensembl_release import EnsemblRelease
-from gtf import GTF
-from locus import Locus
-from compute_cache import clear_cached_objects
-from reference_transcripts import ReferenceTranscripts
+from .ensembl_release import EnsemblRelease
```
iskandr: I had to look around for why they disable implicit relative imports. I guess the logic holds up but it's still an annoying change:

http://programmers.stackexchange.com/questions/159503/whats-wrong-with-relative-imports-in-python

pyensembl/compute_cache.py
iskandr: I think the correct thing to here is to try to `import cPickle as pickle`, and if that fails with an `ImportError` then `import pickle` (and, since this will look insane, comment that we're doing it because GvR made us).

pyensembl/gtf.py
```diff
@@ -144,7 +143,7 @@ def dataframe(self, contig=None, feature=None, strand=None):
         if strand:
             strand = normalize_strand(strand)
 
-        if not isinstance(feature, (NoneType, str, unicode)):
+        if feature is not None and not isinstance(feature, str):
```
iskandr: I guess we're bailing on the possibility of using a unicode string in Python 2.7.x?

pyensembl/locus.py
```diff
@@ -1,10 +1,10 @@
 def normalize_chromosome(c):
-    if isinstance(c, (int, long)):
+    if isinstance(c, int):
```
iskandr: Again, this changes the meaning a little bit for Python 2.7.x. I guess it's fine, since getting a `long` for a chromosome seems like an edge case but it does take us even further from dynamic typing.

pyensembl/reference_transcripts.py
```diff
@@ -115,6 +115,6 @@ def __contains__(self, transcript_id):
         # the pyfaidx __contains__ method requires an expensive list traversal
         # so cache the keys as a set for faster membership checks
         if self._fasta_keys is None:
-            keys = self.fasta_dictionary.keys()
+            keys = list(self.fasta_dictionary.keys())
```
iskandr: This annoys me as "wasteful" but I guess copying a list of transcript IDs won't take more than 10-30ms.

pyensembl/compute_cache.py
```diff
@@ -36,7 +36,7 @@ def clear_cached_objects():
     _memory_cache.clear()
 
 def _read_csv(csv_path):
-    print "Reading Dataframe from %s" % csv_path
+    print("Reading Dataframe from %s" % csv_path)
```
iskandr: Won't work in Python 2.7.x without `from __future__ import print_function`

test/test_exon_object.py
```diff
@@ -97,7 +97,7 @@ def test_partial_start_codons():
     # Exon ENSE00003718948 has a start codon immediately before the
     # exon's start position on the forward strand
     exon = ensembl77.exon_by_id("ENSE00003718948")
-    for start in exon.start_codon_positions:
+    for (start, end) in exon.start_codon_positions:
```
iskandr: Good catch

test/test_gene_names.py
```diff
@@ -20,9 +20,9 @@ def test_all_gene_names(ensembl):
     SMAD4, HSP90AA1, TP53, ERBB2
     """
     gene_names = ensembl.gene_names()
-    print type(gene_names)
+    print(type(gene_names))
     for gene_name in KNOWN_GENE_NAMES:
-        print str(gene_name) in set(gene_names)
+        print(str(gene_name) in set(gene_names))
```
iskandr: `from __future__ import print_function`?

pyensembl/compute_cache.py
```diff
@@ -46,7 +46,7 @@ def _read_csv(csv_path):
     return df
 
 def _write_csv(df, csv_path):
-    print "Saving DataFrame to %s" % csv_path
+    print("Saving DataFrame to %s" % csv_path)
```
iskandr: `from __future__ import print_function`?

iskandr:
Other than the lack of `import print_function` and making sure we're using `cPickle` in Python 2.7.x, LGTM

pyensembl/compute_cache.py
timodonnell: Wow, good catch! Years ago, someone told me that pickle was aliased to cPickle starting in 2.7, an assertion that I never bothered to check, and is apparently wrong. I've been using pickle in so many places...

Fixed according to your suggestion.

pyensembl/gtf.py
timodonnell: yeah, I think that's definitely the simplest thing if it's acceptable

pyensembl/locus.py
timodonnell: Yeah, this was another one I wasn't thrilled about. But, an int in Python 2 is implement as a C "long", which on any modern platform is 64 bits (and guaranteed to be at least 32 bits), so it's really pretty unlikely to need a Python long here.

timodonnell:
I agree that the unicode conversion thing is insanely hacky. i'll try something better

timodonnell:
ok, please take a look at this version. By putting the type checking in a helper module, we now maintain support for unicode in Python 2, and I think made the code more readable as well

iskandr:
Cool, factoring out type checking was on my mental TODO list, thanks for doing it. One possible problem: does the name `assert_string` imply you're actually using an assert statement under the hood? Is there a name that slightly more clear that the raised error won't be an assertion error and that the checks can never be stripped out?

timodonnell:
that's a reasonable point. how about require_string?

iskandr:
Sounds good to me.

timodonnell:
done

PR #36, opened 2015-02-16
don't use sequence to determine length, may not always be available

iskandr:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/36)


timodonnell:
LGTM

PR #39, opened 2015-02-18
Search over loci

iskandr:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/39)


timodonnell:
LGTM

timodonnell:
Actually, I don't see a search module. Was that mistakenly omitted, or am I confused?

iskandr:
Good catch, added it in.

PR #43, opened 2015-02-20
Add two more functions to type_checks.py. These aren't currently used in...

timodonnell:
... pyensembl itself, but are useful in code that uses pyensembl.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/43)


timodonnell:
Motivation for this was hitting a confusing issue in varcode due to a missing type check (VariantCollection.variants isn't checked to be anything), and then added the needed method here to make that type check possible.

iskandr:
Lgtm, if we keep adding to this module we should eventually factor it out into a separate package/repo.

Issue #2, opened 2014-11-21
Use multiple database tables


iskandr:
Instead of putting all the entries in a single "ensembl" table, it would be clearer and faster to have a distinct table for each feature type:
```
exon              1306656
CDS                791856
UTR                304070
transcript         215170
stop_codon          73411
start_codon         73358
gene                63677
Selenocysteine        114
```

Issue #1, opened 2014-11-12
db_from_dataframe() got an unexpected keyword argument 'db_filename'


arahuja:
``` python
In [4]: gene_names = data.gene_names_at_locus(contig_name=12, position=25107806)
INFO:root:Fetching Homo_sapiens.GRCh38.77.gtf.gz from ftp://ftp.ensembl.org/pub/release-77/gtf/homo_sapiens/Homo_sapiens.GRCh38.77.gtf.gz
INFO:root:Downloading ftp://ftp.ensembl.org/pub/release-77/gtf/homo_sapiens/Homo_sapiens.GRCh38.77.gtf.gz
Reading GTF /Users/arahuja/Library/Caches/ensembl/Homo_sapiens.GRCh38.77.gtf.gz into DataFrame
/Users/arahuja/anaconda/lib/python2.7/site-packages/pandas/io/parsers.py:1130: DtypeWarning: Columns (0,3,4) have mixed types. Specify dtype option on import or set low_memory=False.
  data = self._reader.read(nrows)
Extracting attributes for 2672006 entries in GTF DataFrame
Adding attribute columns: ['gene_id', 'gene_version', 'gene_name', 'gene_source', 'gene_biotype', 'transcript_id', 'transcript_version', 'transcript_name', 'transcript_source', 'transcript_biotype', 'exon_number', 'exon_id', 'exon_version', 'tag', 'ccds_id', 'protein_id', 'protein_version']
Saving DataFrame to /Users/arahuja/Library/Caches/ensembl/Homo_sapiens.GRCh38.77.expanded.csv
Creating database: /Users/arahuja/Library/Caches/ensembl/Homo_sapiens.GRCh38.77.db

...
TypeError: db_from_dataframe() got an unexpected keyword argument 'db_filename'
```

Will take a look at a fix for this if I have some time

PR #44, opened 2015-02-23
Adding Travis with pylint

tavinathanson:
I figure we can work out how to make this faster later?

(Also, `pylint` fails on a whole bunch of stuff, but I'm thinking I can address that in a follow-up PR.)

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/44)

### DIFF
--- a/pyensembl/ensembl_release.py
+++ b/pyensembl/ensembl_release.py
@@ -60,8 +60,8 @@ class EnsemblRelease(object):
         is assumed to be some view of the this release's data and thus
         safe to delete.
         """
-        base = self.gtf.base_filename()
-        dirpath = self.local_gtf_dir()
+        base = self.gtf.local_filename()
+        dirpath = self.gtf.local_gtf_path()
         for path in glob(join(dirpath, base + "*")):
             logging.info("Deleting cached file %s", path)
             remove(path)
@@ -374,7 +374,7 @@ class EnsemblRelease(object):
         return self._query_gene_name(
             "transcript_name", transcript_name, 'transcript')
 
-    def gene_name_of_exon_id(self, transcript_id):
+    def gene_name_of_exon_id(self, exon_id):
         return self._query_gene_name("exon_id", exon_id, 'exon')
 
     ###################################################
@@ -382,6 +382,16 @@ class EnsemblRelease(object):
     #             Gene IDs
     #
     ###################################################
+
+    def _query_gene_ids(self, property_name, value, feature='gene'):
+        results = self.db.query(
+            select_column_names=['gene_id'],
+            filter_column=property_name,
+            filter_value=value,
+            feature=feature,
+            distinct=True,
+            required=True)
+        return [str(result[0]) for result in results if result[0]]
 
     def gene_ids(self, contig=None, strand=None):
         """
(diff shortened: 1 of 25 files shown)


iskandr:
I think that a lot of the Travis time will be spent (1) downloading Ensembl
data (2) parsing Ensembl GTFs and (3) indexing the constructed database.

I'm going to fix #3 this week, have some ideas for fixing #2 but we need to
look at getting Travis to cache downloaded data to get the time for #1 down.

tavinathanson:
All makes sense. That said, slow Travis seems better than no Travis for now.

setup.py
```diff
@@ -50,8 +50,8 @@
             'numpy>=1.7',
             'pandas>=0.13.1',
             'datacache>=0.4.2',
-	    'pyfaidx',
-	    'memoized-property',
+            'pyfaidx>=0.3.4',
+            'memoized-property>=1.0.2',
```
iskandr: Instead of duplicating the requirements, can we parse requirements.txt in setup.py? 

Example: http://stackoverflow.com/a/16624700/481326

tavinathanson: Good idea!

iskandr:
Agreed!

setup.py
tavinathanson: (Note that there isn't 100% parity at the moment, since `nose` and `pylint` don't seem necessary when a package is `pip` installed -- but I think it's fine to include them.)

iskandr: re: `nose`, do we install our tests with the package?

tavinathanson: What do you mean?

iskandr: When someone runs `pip install pyensembl`, do they get a copy of our unit tests?

tavinathanson: Not currently: do you think they should?

iskandr: Yeah, especially in a library that depends so much on external resources,
it might be nice to trouble shoot a local problem by seeing what unit tests
fail.

tavinathanson: :+1:

tavinathanson:
I'm going to add PyPi deployment, too: http://docs.travis-ci.com/user/deployment/pypi/

Make sense?

iskandr:
Awesome! Deploying tagged commits seems way nicer than manually submitting
to PyPI. However, will our credentials end up in the repo in plain text?

tavinathanson:
It supports password encryption :)

iskandr:
@tavinathanson I'd merge this if I could

tavinathanson:
@iskandr I rebased and added PyPi support (which I doubt will work flawlessly on the first try, but let's see)

tavinathanson:
(Oh, and parsing of `requirements.txt`)

tavinathanson:
D'oh, I should `raise` if there's no `requirements.txt`

iskandr:
Want to bump the version and tag the branch? It'll be our first experiment in auto-PyPI.

tavinathanson:
@iskandr I created a tag and did `git push --tags`, is that all I need to do?

tavinathanson:
@iskandr Actually, it won't be our first experiment because the linter will fail. Guess I should address the linting errors now...

pyensembl/ensembl_release.py
```diff
@@ -495,12 +505,13 @@ def transcript_name_of_transcript_id(self, transcript_id):
...
 
-    def _query_transcript_ids(self, property_name, value):
+    def _query_transcript_ids(self, property_name, value,
+                              feature='transcript'):
         results = self.db.query(
             select_column_names=['transcript_id'],
             filter_column=property_name,
             filter_value=value,
-            feature='transcript',
+            feature=feature,
```
iskandr: This is probably a needless generalization, since we're looking up distinct transcript IDs and I can't think of any feature we would to find those on other than transcripts. I guess this way you can ask "find the transcript ID of any exon which matches my filter criteria..."

pyensembl/ensembl_release.py
```diff
@@ -60,8 +60,8 @@ def _delete_cached_files(self):
         is assumed to be some view of the this release's data and thus
         safe to delete.
         """
-        base = self.gtf.base_filename()
-        dirpath = self.local_gtf_dir()
+        base = self.gtf.local_filename()
```
iskandr: I think `base_filename` is actually what you want here, since all the locally cached files look like `base_filename + "_something_extra.csv"`

setup.py
```diff
@@ -14,16 +14,27 @@
...
-  pass
+    print("Conversion of long_description from MD to reStructuredText failed...")
+    pass
+
+
+requires_list = []
+try:
+    from pip.req import parse_requirements
+    parsed_reqs = parse_requirements('requirements.txt')
+    requires_list = [str(req.req) for req in parsed_reqs]
```
iskandr: I guess we'll change this later to use your setup helpers library.

iskandr:
Small comments but otherwise LGTM

PR #46, opened 2015-02-25
fixed mixing up the local and remote filenames of GTF files

iskandr:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/46)


tavinathanson:
Looks good to me

PR #52, opened 2015-03-02
Faster database schema

iskandr:
The database schema was previous just a single table called `ensembl`, which required every query to specify e.g. `feature='transcript'`. This branch splits the schema into multiple tables (one for genes, one for transcripts, &c). This allows `transcript_id` and `gene_id` to be primary keys of their respective tables, and speeds up searches for a particular ID by reducing the number of rows which must be inspected. 

This branch relies on the datacache branch https://github.com/hammerlab/datacache/pull/16

The impact on performance is ~10x faster for gene lookups, ~3x faster for transcript lookups and confusingly ~2x slower for exon lookups. Since the latter are rarer, this is a net win but it's still worth figuring out how/why exon queries became slower. 

Other changes:
- added license headers to several files
- deleted `type_checks` module, using the `typechecks` pip package instead
- parsing requirements in setup.py from requirements.txt

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/52)


hammer:
I wouldn't spend too much time tuning the database performance with sqlite. At some point we'll want to move to a real database and then we can start digging in to performance characteristics of the workload.

iskandr:
I agree, figuring out the performance of exon lookups is a very low
priority for me. As for this PR, a schema with one table per feature
(rather than a big monolithic table) is what we'll want regardless of the
backend.

tavinathanson:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pyensembl/52)

Reviewed files:
- pyensembl/__init__.py @ r1
- pyensembl/biotypes.py @ r1
- pyensembl/common.py @ r1
- pyensembl/compute_cache.py @ r1
- pyensembl/database.py @ r1
- pyensembl/ensembl_release.py @ r1
- pyensembl/exon.py @ r1
- pyensembl/gene.py @ r1
- pyensembl/gtf.py @ r1
- pyensembl/gtf_parsing.py @ r1
- pyensembl/locus.py @ r1
- pyensembl/reference_transcripts.py @ r1
- pyensembl/release_info.py @ r1
- pyensembl/search.py @ r1
- pyensembl/shell.py @ r1
- pyensembl/species.py @ r1
- pyensembl/transcript.py @ r1
- pyensembl/type_checks.py @ r1
- pyensembl/url_templates.py @ r1
- requirements.txt @ r1
- setup.py @ r1
- test/test_timings.py @ r1





---

<sup>**[pyensembl/database.py, line 27 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/52#-JjWNE1-VKHVCoqrzY96)** ([raw file](https://github.com/hammerlab/pyensembl/blob/6c4773172ee78c1152958b3fb0575e89eda3aec0/pyensembl/database.py#L27)):</sup>
Should this also go in the README?

---

<sup>**[pyensembl/database.py, line 46 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/52#-JjWOD2Zzs3s7BD4UTXb)** ([raw file](https://github.com/hammerlab/pyensembl/blob/6c4773172ee78c1152958b3fb0575e89eda3aec0/pyensembl/database.py#L46)):</sup>
:+1:

---

<sup>**[pyensembl/database.py, line 61 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/52#-JjRIpSTr3tyoZX0bfQE)** ([raw file](https://github.com/hammerlab/pyensembl/blob/6c4773172ee78c1152958b3fb0575e89eda3aec0/pyensembl/database.py#L61)):</sup>
What happened to seqname?

---

<sup>**[pyensembl/database.py, line 103 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/52#-JjWQ_aYLGEw0Vtx8O-i)** ([raw file](https://github.com/hammerlab/pyensembl/blob/6c4773172ee78c1152958b3fb0575e89eda3aec0/pyensembl/database.py#L103)):</sup>
This took me a bit to make sense of (each table getting the same indices), but I get it. Maybe a brief explanation of why: "because the tables, regardless of the feature they contain, should be filterable by gene_id, transcript_id, etc."

---

<sup>**[pyensembl/database.py, line 113 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/52#-JjWR6tTbwLdLPKoAb6q)** ([raw file](https://github.com/hammerlab/pyensembl/blob/6c4773172ee78c1152958b3fb0575e89eda3aec0/pyensembl/database.py#L113)):</sup>
I think i'd prefer this little "%s_id" line of logic pulled out, e.g. get_primary_key(feature), with a docstring clarification. The simple assumption of being able to add _id worries me, but you know the Ensembl data better than I do.

Where does protein_id fit in? In https://github.com/hammerlab/pyensembl/pull/44/files, I needed to set feature='CDS' to grab protein_id, but I very well might have been doing things wrong.

---

<sup>**[pyensembl/database.py, line 117 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/52#-JjWSmY0KioVeImgILiA)** ([raw file](https://github.com/hammerlab/pyensembl/blob/6c4773172ee78c1152958b3fb0575e89eda3aec0/pyensembl/database.py#L117)):</sup>
I think checking for nulls and uniqueness should also be factored out into an is_valid_primary_key method.

---

<sup>**[pyensembl/database.py, line 171 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/52#-JjWTOcHmvdOf6lM-enW)** ([raw file](https://github.com/hammerlab/pyensembl/blob/6c4773172ee78c1152958b3fb0575e89eda3aec0/pyensembl/database.py#L171)):</sup>
This should be changed in a few spots, as this error is thrown in a few spots. (Wasn't sure of the best way to centralize message, though that'd be great.)

Also, are you preparing to move beyond Ensembl?

---

<sup>**[setup.py, line 33 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/52#-JjWZOCoXDkxG6n8YzML)** ([raw file](https://github.com/hammerlab/pyensembl/blob/6c4773172ee78c1152958b3fb0575e89eda3aec0/setup.py#L33)):</sup>
See my comments in the datacache PR, and also https://github.com/hammerlab/pyensembl/blob/travis/setup.py: I'd prefer to do it that way, and I also think it'd be good to share this logic between these different repos. (The logic sharing can be a separate PR.)

---

iskandr:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pyensembl/52)







---

<sup>**[pyensembl/database.py, line 27 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/52#-JjWNE1-VKHVCoqrzY96)** ([raw file](https://github.com/hammerlab/pyensembl/blob/6c4773172ee78c1152958b3fb0575e89eda3aec0/pyensembl/database.py#L27)):</sup>
I don't think so, it's an implementation detail that only matters when you're modifying the contents of this module.

---

<sup>**[pyensembl/database.py, line 61 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/52#-JjRIpSTr3tyoZX0bfQE)** ([raw file](https://github.com/hammerlab/pyensembl/blob/6c4773172ee78c1152958b3fb0575e89eda3aec0/pyensembl/database.py#L61)):</sup>
I realized it's a silly database index since the number of matching results is still gigantic, and I'm not often asking for "all the features on this contig". Doesn't seem worth the extra time spent indexing upon install.

---

<sup>**[pyensembl/database.py, line 103 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/52#-JjWQ_aYLGEw0Vtx8O-i)** ([raw file](https://github.com/hammerlab/pyensembl/blob/6c4773172ee78c1152958b3fb0575e89eda3aec0/pyensembl/database.py#L103)):</sup>
There are some situations when a feature totally lacks non-null values for some index column(s), so creating that index is a waste of time. Handled via `_feature_indices` helper. 

---

<sup>**[pyensembl/database.py, line 113 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/52#-JjWR6tTbwLdLPKoAb6q)** ([raw file](https://github.com/hammerlab/pyensembl/blob/6c4773172ee78c1152958b3fb0575e89eda3aec0/pyensembl/database.py#L113)):</sup>
You're right, I didn't realize that protein_id is effectively acting a primary key for CDS. 

I like your idea of moving the tricky stuff into helpers, I split out a lot of the logic inside this loop into `_get_primary_key` and `_feature_indices`. 

---

<sup>**[pyensembl/database.py, line 117 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/52#-JjWSmY0KioVeImgILiA)** ([raw file](https://github.com/hammerlab/pyensembl/blob/6c4773172ee78c1152958b3fb0575e89eda3aec0/pyensembl/database.py#L117)):</sup>
I disagree, I find that when I'm reading code, one liners with clear logic are actually easier to understand than helper methods. I would still wonder how we know whether a primary key is "valid" or what that means, but here it's spelled out without much fuss. 

---

<sup>**[pyensembl/database.py, line 171 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/52#-JjWTOcHmvdOf6lM-enW)** ([raw file](https://github.com/hammerlab/pyensembl/blob/6c4773172ee78c1152958b3fb0575e89eda3aec0/pyensembl/database.py#L171)):</sup>
Maybe we can make an `error_messages` module? That seems over the top. I'm not really sure how to centralize this either. 

Re; "move beyond Ensembl";  you mean with the "Ensembl" -> "Genome" change? Nope, I just though that it's clearer to call it "genome annotation data", since, well, we're downloading annotations over a genome and from the context it should be pretty clear we're doing that through Ensembl.

---

<sup>**[setup.py, line 33 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/52#-JjWZOCoXDkxG6n8YzML)** ([raw file](https://github.com/hammerlab/pyensembl/blob/6c4773172ee78c1152958b3fb0575e89eda3aec0/setup.py#L33)):</sup>
See: https://github.com/machinalis/iepy/issues/63

(pip doesn't have a stable external API)

I agree with your idea in the other PR, we should make a little library of setup helpers. 

---

iskandr:
(rebased off master)

tavinathanson:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pyensembl/52)

Reviewed files:
- pyensembl/gtf.py @ r1
- pyensembl/gtf_parsing.py @ r1
- pyensembl/locus.py @ r1
- pyensembl/reference_transcripts.py @ r1
- pyensembl/release_info.py @ r1
- pyensembl/search.py @ r1
- pyensembl/shell.py @ r1

---

Looks good to me! Sorry for the delay.

---

<sup>**[pyensembl/database.py, line 129 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/52#-JjWSmY0KioVeImgILiA)** ([raw file](https://github.com/hammerlab/pyensembl/blob/6c4773172ee78c1152958b3fb0575e89eda3aec0/pyensembl/database.py#L129)):</sup>
Ah, I wasn't necessarily saying that get_primary_key needed an additional helper method, so your factoring out the primary key logic meets my needs!

---

iskandr:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pyensembl/52)



---

No worries!

---

PR #53, opened 2015-03-05
Fix missing type checks module

iskandr:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/53)


danvk:
LGTM

Is it usually to define `__version__` somewhere in your Python code, too?

iskandr:
Know of a good way to define __version__ without duplicating it in setup.py?

danvk:
I haven't the foggiest.

Issue #58, opened 2015-03-09
error in database.py: Column 'protein_id' can't be primary key of table 'CDS' because it contains repeated values


timodonnell:
This is while running `pyensembl 75 install`

```
Traceback (most recent call last):
  File "/Users/tim/sinai/git/datacache/venv/bin/pyensembl", line 9, in <module>
    load_entry_point('pyensembl==0.5.12', 'console_scripts', 'pyensembl')()
  File "/Users/tim/sinai/git/pyensembl/pyensembl/shell.py", line 59, in run
    data.install()
  File "/Users/tim/sinai/git/pyensembl/pyensembl/ensembl_release.py", line 152, in install
    self._index(force=False)
  File "/Users/tim/sinai/git/pyensembl/pyensembl/ensembl_release.py", line 163, in _index
    if self.db.create(force=force):
  File "/Users/tim/sinai/git/pyensembl/pyensembl/database.py", line 515, in create
    self._create_database(force=force)
  File "/Users/tim/sinai/git/pyensembl/pyensembl/database.py", line 170, in _create_database
    primary_key = self._get_primary_key(feature, df_subset)
  File "/Users/tim/sinai/git/pyensembl/pyensembl/database.py", line 127, in _get_primary_key
    primary_key, feature_name))
ValueError: Column 'protein_id' can't be primary key of table 'CDS' because it contains repeated values

timodonnell:
This occurs for both releases 75 and 77

PR #60, opened 2015-03-09
don't use protein_id as primary key of CDS table

iskandr:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/60)


iskandr:
Fixes https://github.com/hammerlab/pyensembl/issues/58

PR #83, opened 2015-03-30
Make all PyEnsembl objects hashable

iskandr:
Implementation of __hash__ needed for memoization (maybe only in Python 3?)

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/83)

### DIFF
--- a/pyensembl/database.py
+++ b/pyensembl/database.py
@@ -45,6 +45,9 @@ class Database(object):
     def __eq__(self, other):
         return isinstance(other, Database) and self.gtf == other.gtf
 
+    def __hash__(self):
+        return hash(self.gtf)
+
     def local_db_filename(self):
         base = self.gtf.base_filename()
         return base + ".db"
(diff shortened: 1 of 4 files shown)


pyensembl/database.py
```diff
@@ -45,6 +45,9 @@ def __init__(self, gtf, auto_download=False):
     def __eq__(self, other):
         return isinstance(other, Database) and self.gtf == other.gtf
 
+    def __hash__(self):
```
timodonnell: Sorry that this is unrelated to this PR, but can we add to class docstring (or somewhere else) what `gtf` is? Or perhaps rename that to not be an abbreviation

timodonnell: Ah ok, it's a file format. Would be good to clear up what the type of`gtf` instance variable is supposed to be though.

timodonnell:
LGTM

PR #86, opened 2015-04-22
added prebuilt EnsemblRelease objects for major human genomes

iskandr:
* Added cached EnsemblRelease construction helper
* Made EnsemblRelease objects for major genomes available in `pyensembl` (so you don't have to keep initializing new objects yourself).

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/86)


pyensembl/__init__.py
```diff
@@ -19,6 +19,26 @@
...
+    """
+    version = check_release_number(version)
+    key = version, auto_download
+    if key not in _cache:
+        ensembl = EnsemblRelease(version, auto_download=auto_download)
+        _cache[key] = ensembl
+    return _cache[key]
+
+ensembl_grch36 = ensembl54 = cached_release(54)  # last release for GRCh36/hg18
+ensembl_grch37 = ensembl75 = cached_release(75)  # last release for GRCh37/hg19
```
arahuja: Does this end up pre-fetching all of the major ensembl releases?  Does this happen on install then?

iskandr: No, it will only download anything if you start using the methods on these objects.

PR #87, opened 2015-04-22
Added "gene" property to Transcript objects which gives access to the parent Gene

iskandr:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/87)

### DIFF
--- a/pyensembl/database.py
+++ b/pyensembl/database.py
@@ -387,6 +387,7 @@ class Database(object):
 
         return results
 
+    @memoize
     def query(
             self,
             select_column_names,
@@ -411,6 +412,7 @@ class Database(object):
         return self.run_sql_query(
             sql, required=required, query_params=query_params)
 
+    @memoize
     def query_one(
             self,
             select_column_names,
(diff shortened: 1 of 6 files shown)


arahuja:
Looks like there are CI issues - is that expected? 

I can't comment on this line, but in `__init__`:
```
     def __init__(
             self,
             transcript_id,
             transcript_name,
             contig,
             start,
             end,
             strand,
             biotype,
             gene_id,
             gene_name,
```

`Transcript probably doesn't need `gene_name` anymore since you can get the from `gene`, or is that costly?

iskandr:
Yeah, "fixing Travis" has been on the TODO list for a long time. I think @tavinathanson has some idea for it but it's not done yet.

tavinathanson:
Yeah, I started on the GTF minification but haven't looked at that for a few weeks. Just disabled Travis CI so we stop getting errors.

iskandr:
@arahuja Removed `gene_name` from the Transcript constructor, you're right that I can just use the `Gene` object instead.

PR #88, opened 2015-04-23
print biotype as part of Transcript and Gene

iskandr:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/88)

### DIFF
--- a/pyensembl/common.py
+++ b/pyensembl/common.py
@@ -24,19 +24,21 @@ def _memoize_cache_key(args, kwargs):
     Expects that all arguments to a memoized function are either hashable
     or can be uniquely identified from type(arg) and repr(arg).
     """
-    cache_key = args + tuple(sorted(kwargs.items()))
+    cache_key_list = []
 
-    try:
-        # if any element of the cache key isn't hashable then we switch
-        # to using the types and string representations of
-        # all the elements in the cache key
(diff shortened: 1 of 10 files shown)


pyensembl/common.py
```diff
@@ -24,19 +24,21 @@ def _memoize_cache_key(args, kwargs):
...
-        hash(cache_key)
-    except TypeError:
-        cache_key = tuple(
-            (type(key_element), repr(key_element))
-            for key_element in cache_key
-        )
-    return cache_key
+    cache_key_list = []
+
+    # hack to get around the unhashability of lists,
```
tavinathanson: Cool, like we discussed in another PR, I like this more despite the hack.

pyensembl/database.py
```diff
@@ -55,7 +55,7 @@ def __init__(self, gtf, auto_download=False):
         self.logger.setLevel(logging.INFO)
 
     def __eq__(self, other):
-        return isinstance(other, Database) and self.gtf == other.gtf
+        return other.__class__ is Database and self.gtf == other.gtf
```
tavinathanson: Why the move from `isinstance` to `__class__`?

tavinathanson:
Overall: looks good after the fact + 1 question

PR #91, opened 2015-06-01
Allow pyensembl to use any DB-API backend

tavinathanson:
cc @arahuja 

Intimately related to https://github.com/hammerlab/datacache/pull/25; see that PR for more details.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/91)

### DIFF
--- a/pyensembl/database.py
+++ b/pyensembl/database.py
@@ -15,8 +15,10 @@
 from __future__ import print_function, division, absolute_import
 
 import logging
-from os.path import join, exists
-import sqlite3
+from os.path import join
+from sqlalchemy import MetaData
+from sqlalchemy.sql import text
+from sqlalchemy.exc import OperationalError
 
 import datacache
 from typechecks import require_integer, require_string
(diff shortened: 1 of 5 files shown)


pyensembl/database.py
```diff
@@ -203,30 +233,30 @@ def _create_database(self, force=False):
...
     @property
     def connection(self):
         """
-        Return the sqlite3 database for this Ensembl release
-        (download and/or construct it if necessary, if auto_download
-        is on). As a side effect, stores the database connection in
-        self._connection.
+        Return the database for this Ensembl release (download and/or 
+        construct it if necessary, if auto_download is on). As a side 
+        effect, stores the database connection in self._connection.
```
timodonnell: I think adding explicitly the return type to the docstring (is it a sql alchemy connection, or something else?) would make this clearer

timodonnell:
Thanks for doing this, @tavinathanson and sorry for my slow turnaround. Overall looks good to me. 

Especially since this code is generating SQL queries directly instead of using the sqlalchemy expression library (a choice I generally agree with -- I've found the expression library to often be pretty troublesome), I think we need to have some sort of test that this works on postgres and whatever other databases we're targeting. Is there a way to add an easy way for the user to configure the unit tests to run on a local postgres database? (Apologies if you've already done this and I missed it. It might just need to be documented more obviously.)

tavinathanson:
This is way out of date, so closing it for now.

PR #99, opened 2015-07-16
Handle arbitrary GTFs and FASTA files, both local and remote

tavinathanson:
Summary:

* New `Genome` class replaces `EnsemblRelease` class.
* New `GenomeSource` class specifies URLs to GTFs and FASTAs, extracted into its own class to prevent argument overload in `Genome` and to create an easy object to pass around for generating python/console install strings (e.g. `pyensembl install --release 77` and `pyensembl install --gtf_url blah`). `EnsemblReleaseSource` extends `GenomeSource` to handle Ensembl URL creation and install strings.
* `GenomeSource` accepts both local file paths and remote URLs. If the latter, datacache is bypassed.
* Added logic to `gene.py`, `transcript.py`, `gtf_parsing.py` and `genome.py` to allow for missing gene names, transcript names and biotypes in GTF files. (I noticed that this can happen with UCSC GTF files from genome.ucsc.edu/cgi-bin/hgTables, such as the one included in this PR as a test.) See this commit: https://github.com/hammerlab/pyensembl/commit/a2321156ccfb7e682750e8fc555e5cb61e7e9bf0
* Added one basic test for a UCSC GTF file.
* Modified `shell.py` to allow for specific URLs in addition to release numbers.

This needs more testing, but I want to get the PR started in case you think the overall strategy is problematic.

Also TODO, either in this PR or a follow-up:

* Error gracefully when a transcript/protein sequence is requested, yet no URL was provided.
* Error gracefully when a user tries to get a gene name when the DB has no gene names (and other issues like that).

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/99)


tavinathanson:
Since this view doesn't show what's changed with `Genome` vs. `EnsemblRelease`, here's a diff of old `EnsemblRelease` vs. new `Genome`: https://gist.github.com/tavinathanson/eb786789a022153d9470

tavinathanson:
@iskandr Another thought: I'm thinking a better API would probably involve saving a collection of URLs/paths as a new "release" in the DB, rather than always requiring each path to be specified. This would also more easily allow topiary to refer to that collection of paths.

For example:

`pyensembl add-genome --name "mouse_81" --gtf-path <URL> --transcript-fasta-path <URL>`
and then
`pyensembl install "mouse_81"`
`topiary --pyensembl-genome "mouse_81"`

`mouse_81` could be saved to the database, mapped to those paths.

I think I'll save that for a follow-up PR. Thoughts?

pyensembl/gtf_parsing.py
```diff
@@ -339,8 +353,8 @@ def load_gtf_as_dataframe(filename):
         logging.info("Creating entries for feature='transcript'")
         df = reconstruct_transcript_rows(df)
 
-    if 'exon_id' not in df:
-        logging.info("Creating 'exon_id' column")
-        df = reconstruct_exon_id_column(df)
+    #if 'exon_id' not in df:
```
tavinathanson: Not sure how this ended up here! Will uncomment.

pyensembl/biotypes.py
```diff
@@ -244,6 +244,8 @@
     'IG_J_pseudogene',
     'IG_pseudogene',
     'IG_V_pseudogene',
+    # Found in ftp://ftp.ensembl.org/pub/release-81/gtf/mus_musculus
+    'IG_D_pseudogene',
```
iskandr: Any ideas for automatically generating this biotypes list? I'm terrified of how perpetually out of date this list will always be.

pyensembl/database.py
```diff
@@ -74,7 +74,7 @@ def _all_possible_indices(self, column_names):
         Create list of tuples containing all possible index groups
         we might want to create over tables in this database.
 
-        If a release is missing some column we want to index on,
+        If a genomic database is missing some column we want to index on,
```
iskandr: What do you think of using "genome annotation" to talk about the set of annotation data instead of "genomic database" to avoid confusion with the actual sqlite database we're creating?

pyensembl/biotypes.py
tavinathanson: No great ideas at the moment, but this "fix" concerns me too :\

pyensembl/database.py
```diff
@@ -223,7 +224,7 @@ def _connect_if_exists(self):
     @property
     def connection(self):
         """
-        Return the sqlite3 database for this Ensembl release
+        Return the sqlite3 database for this Genomic database
```
iskandr: "Genomic" is unexpectedly capitalized (or maybe should be "genome annotation" or "genome annotation dataset")

pyensembl/ensembl_release_source.py
```diff
@@ -0,0 +1,77 @@
...
+from os.path import split
+
+from .genome_source import GenomeSource
+from .release_info import MAX_ENSEMBL_RELEASE
+from .url_templates import ENSEMBL_FTP_SERVER, make_gtf_url, make_fasta_url
+
+
+class EnsemblReleaseSource(GenomeSource):
+    """
+    Represents the source (URLs) of an Ensembl release. The superclass
```
iskandr: Which superclass? `GenomeSource`? Why talk about it in a derived class?

pyensembl/database.py
tavinathanson: I tried, but didn't quite succeed, to use consistent verbiage throughout: mostly because I was conflicted about it!

How do I refer to the collection of stuff that includes GTF and FASTA? Annotation seems good for just the GTF part?

pyensembl/database.py
tavinathanson: Will fix

pyensembl/genome.py
```diff
@@ -0,0 +1,848 @@
...
+            self.reference_name,
+            self.only_human,
+            self.genome_source)
+
+    def __repr__(self):
+        return str(self)
+
+    def __eq__(self, other):
+        return (
+            other.__class__ is Genome and
```
iskandr: Though experiment: what happens if you create a `Genome` with all the GTF/FASTA files of an Ensembl release? In what ways will the two objects differ (i.e. why aren't they equal?)

pyensembl/ensembl_release_source.py
tavinathanson: Yeah, `GenomeSource`. I was trying to clarify how it relates to the superclass, but am fine with just removing that line.

pyensembl/genome.py
```diff
@@ -0,0 +1,848 @@
...
+        transcript doesn't have cDNA sequence.
+        """
+        assert self.protein_sequences, (
+            "This genome source does not include protein FASTA data: %s"
+            % self.genome_source)
+        if self.only_human:
+            require_human_protein_id(protein_id)
+        return self.protein_sequences.get(protein_id)
+
+    def download(self, force=True):
```
iskandr: Can you tell me more about why you kept `download` as part of `Genome` instead of moving it to `GenomeSource`?

pyensembl/genome.py
tavinathanson: Totally valid thought experiment. I've been wondering about that: also in the case of using `Genome` for mice rather than expanding `EnsemblRelease` to accommodate mice.

Maybe `EnsemblRelease.as_genome()` would be helpful here? And then that resultant object would be equal to the other?

pyensembl/genome.py
tavinathanson: That thought literally never crossed my mind. Frankly, I wasn't sure if you'd like the `Genome`/`GenomeSource` split in the first place.

I do think that `download`, `install` and `index` should probably live in the same class. Maybe that's a reason to keep them all in `Genome`?

pyensembl/database.py
iskandr: Yeah, I'm confused about this too. It's also confusing to call the whole bundle a `Genome` when really only the underlying whole genome sequence is, well, a genome. The transcript and protein FASTA files are derived concepts, as well as the GTF describing the coordinates of various genomic features. 

One thought:
*  "genomic feature annotations" == GTF

Other than that, I'm not sure.

pyensembl/ensembl_release_source.py
iskandr: Maybe it's better to actually add more! I'm curious what extra functionality you get from `EnsemblReleaseSource` that isn't in `GenomeSource`.

tavinathanson: In short: specifies what Ensembl URLs actually look like, and overrides the console/python install strings to be Ensembl-specific (i.e. "--release"). Guess I'll just write that!

iskandr:
OK, here's a thought: 

"genome" = "genome annotation" (GTF) + "genome sequence" (FASTA files)

So, we may eventually want to add genome sequences separately from annotations. 

I like your idea below: 

`pyensembl add-genome --name "mouse_81" --gtf-path <URL> --transcript-fasta-path <URL>`

but might want to even extend it to:

`pyensembl add-genome-sequence "mouse" --transcript-fasta-path <URL>`

`pyensembl add-genome --name "mouse_81" --gtf-path <URL> --genomoe-sequence "mouse"`

Agreed that we can figure this out in a later PR.

pyensembl/genome.py
```diff
@@ -0,0 +1,848 @@
...
+        self.name = name
+        self.version = version
+        self.only_human = only_human
+        self.auto_download = auto_download
+
+        # GTF object wraps the source GTF file from which we get
+        # genome annotations. Presents access to each feature
+        # annotations as a pandas.DataFrame.
+        self.gtf = GTF(
+            genome_source,
```
iskandr: Does GTF need all the info from `genome_source`? It seems like the FASTA URLs shouldn't get passed here.

pyensembl/genome.py
iskandr: In what way would `EnsemblRelease.as_genome()` differ from the original ensembl_release (other than its class tag)? I'm trying to understand why `EnsemblRelease` needs to continue existing as its own class.

pyensembl/genome.py
iskandr: ...or maybe move them all to `GenomeSource`? I'm not sure if that's better but I'm having trouble articulating what `GenomeSource` *does* (other than storing some related URLs).

pyensembl/genome_source.py
```diff
@@ -0,0 +1,71 @@
...
+            args.append("%s=\"%s\"" % (name, path))
+        return ", ".join(args)
+
+    def install_string_python(self):
+        return "Genome(GenomeSource(%s)).install()" % self._arg_list_str()
+
+    def fasta_path(self, fasta_type):
+        assert fasta_type in ["transcript", "protein"], \
+            "Invalid FASTA type: %s" % fasta_type
+        if fasta_type == "transcript":
```
iskandr: This is a little funky-- why not have `protein_fasta_path` and `transcript_fasta_path` as distinct methods?

pyensembl/sequence_data.py
```diff
@@ -20,33 +20,42 @@
...
+    Container for reference nucleotide and amino acid sequenes.
 
     Downloads and caches FASTA file, creates database mapping
     identifiers to sequences.
     """
     def __init__(
             self,
-            ensembl_release,
-            fasta_url,
+            genome_source,
```
iskandr: Instead of passing a `genome_source` and `fasta_type`, why not just give the fasta path directly?

tavinathanson: I went back and forth on this. On the one hand, `GenomeSource` specifies the install strings that I use in error messages here, so the class itself is handy. (And whether those install strings, e.g. `pyensembl install --gtf-path` should specify all the GTF/FASTA information, is unclear in my mind.) On the other hand, I agree that the rest of the data isn't really needed. This seemed fine enough, even if not perfect?

pyensembl/genome.py
tavinathanson: Discussed offline: we'll keep `EnsemblRelease` for now. The question of whether to rip it out later will also relate to how we want to represent a `RefSeqRelease` (its own class?).

pyensembl/genome.py
tavinathanson: Discussed offline: `GenomeSource` should either do more or not exist, though we'll keep it for this PR.

It does seem nice to separate `index`/`install`/`download` from all the gene related methods, so in that sense moving them to `GenomeSource` would be cool.

That said, if someone just wants to run `EnsemblRelease(77).install()`, it would be annoying if they actually needed to touch a different class (`EnsemblReleaseSource`) as well.

pyensembl/genome.py
tavinathanson: Discussed offline: `genome_source` is also used for the error string. For now, we'll pass in both the path/URL and the `genome_source`.

pyensembl/genome_source.py
tavinathanson: Discussed offline: this relates to `SequenceData`'s use of `GenomeSourcce` to get the path. Instead, I'll just pass both the path/URL and the `GenomeSource`, and remove this `fasta_path` functionality.

tavinathanson:
"genome" = "genome annotation" (GTF) + "genome sequence" (FASTA files) sounds reasonable for talking about these entities for the time being, even though as we've discussed "genome" is still confusing.

Made https://github.com/hammerlab/pyensembl/issues/100 for `add-genome`

tavinathanson:
TODO before merging, from offline discussion:

- [ ] Add the actual path to `SequenceData` and `GTF`
- [ ] Remove `fasta_path` from `GenomeSource`
- [ ] Change `path` to `path_or_url` in most places
- [ ] Move `local_fasta_filename_func` to `EnsemblReleaseSource`, and local/remote filename logic to `GenomeSource`
- [ ] Rename local/remote to original/cached

tavinathanson:
I believe all code review comments are now addressed.

@iskandr I ended up changing the role of `GenomeSource`: it now represents a single URL or path. It continues to return install messages, but now does other useful things: it handles the original/cached filename transform (like we discussed), and it also handles all the `is_url_format` and local copying logic. It's purely internal, now; I added the GTF/FASTA arguments to `Genome` itself.

It's not perfect, but I think/hope it's moving in the right direction. I won't be able to address further comments until next week. If you think it's ready to merge, feel free!

iskandr:
I'm surprised that a `Genome` now has multiple `GenomeSource` objects for each GTF and FASTA file. I'm also surprised that `EnsemblReleaseSource` survived despite that change in the role of source objects. Still, since you're gone for a week and kept the API backward compatible I'd rather merge this now and then discuss this zoo of objects when you get back.

iskandr:
I tried to run the unit tests and got the following error: 

```
nose.proxy.TypeError: unhashable type: 'GenomeSource'
```

iskandr:
Running the unit tests with Python 2.7 I get different errors:

```
ERROR: test_all_gene_names : Make sure some known gene names such as
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/Cellar/python/2.7.6/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/nose-1.3.6-py2.7.egg/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/iskander/code/pyensembl/test/test_common.py", line 35, in new_test_fn
    test_fn(ensembl)
  File "/Users/iskander/code/pyensembl/test/test_gene_names.py", line 23, in test_all_gene_names
    gene_names = ensembl.gene_names()
  File "/Users/iskander/code/pyensembl/pyensembl/common.py", line 56, in wrapped_fn
    value = fn(*args, **kwargs)
  File "/Users/iskander/code/pyensembl/pyensembl/genome.py", line 519, in gene_names
    strand=strand)
  File "/Users/iskander/code/pyensembl/pyensembl/genome.py", line 218, in all_feature_values
    return cached_object(pickle_path, compute_fn=run_query)
  File "/Users/iskander/code/pyensembl/pyensembl/compute_cache.py", line 118, in cached_object
    obj = pickle.load(f)
ValueError: unsupported pickle protocol: 4
```

iskandr:
Problems fixed, though it seems like the pickling issues may require nuking all `.pickle` files in the cache.

Issue #107, opened 2015-07-31
Small typo in GenomeSource.cached_fasta_path


iskandr:
Looks like a comma morphed itself into a dot. 

```
  File "/Users/iskander/code/varcode/test/test_mouse.py", line 33, in test_specific_variant_mouse
    effects = variant.effects()
  File "/Users/iskander/code/varcode/varcode/common.py", line 47, in wrapped_fn
    memoized_values[key] = fn(*args, **kwargs)
  File "/Users/iskander/code/varcode/varcode/variant.py", line 394, in effects
    effect = self.effect_on_transcript(transcript)
  File "/Users/iskander/code/varcode/varcode/common.py", line 47, in wrapped_fn
    memoized_values[key] = fn(*args, **kwargs)
  File "/Users/iskander/code/varcode/varcode/variant.py", line 429, in effect_on_transcript
    if not transcript.complete:
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/memoized_property.py", line 40, in fget_memoized
    setattr(self, attr_name, fget(self))
  File "/Users/iskander/code/pyensembl/pyensembl/transcript.py", line 379, in complete
    self.coding_sequence is not None and
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/memoized_property.py", line 40, in fget_memoized
    setattr(self, attr_name, fget(self))
  File "/Users/iskander/code/pyensembl/pyensembl/transcript.py", line 425, in coding_sequence
    return self.sequence[start:end + 1]
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/memoized_property.py", line 40, in fget_memoized
    setattr(self, attr_name, fget(self))
  File "/Users/iskander/code/pyensembl/pyensembl/transcript.py", line 389, in sequence
    return self.ensembl.transcript_sequences.get(self.id)
  File "/Users/iskander/code/pyensembl/pyensembl/sequence_data.py", line 236, in get
    fasta_record = self.fasta_dictionary.get(sequence_id)
  File "/Users/iskander/code/pyensembl/pyensembl/sequence_data.py", line 224, in fasta_dictionary
    self.index()
  File "/Users/iskander/code/pyensembl/pyensembl/sequence_data.py", line 191, in index
    if exists(self.cached_database_path):
  File "/Users/iskander/code/pyensembl/pyensembl/sequence_data.py", line 167, in cached_database_path
[...]

Issue #106, opened 2015-07-31
When sequence data is not provided, error gracefully


tavinathanson:
Right now, when sequence data is not provided, we just run into an undefined variable.

For example, in `transcript.py` (stolen from @iskandr):

```
if self.sequence is None:
    raise ValueError("No 3' UTR sequence for %s" % self)
```

tavinathanson:
We should also error gracefully when GTF data is not provided.

tavinathanson:
Fixed by #108

Issue #105, opened 2015-07-31
EnsemblReleaseSource is not used


tavinathanson:
Noticed by @iskandr. This creates the following issues:

* The install string printed in error messages is incorrect (no release number).
* The Ensembl-specific local/remote transform is ignored.

This should be fixed by actually using `EnsemblReleaseSource` internally.

tavinathanson:
This also resulted in a super weird bug, with `varcode` tests complaining about missing FASTA information. Because the logic to add the release number to the FASTA filename is not currently included, the FASTA information for one release can clobber that of another!

Issue #104, opened 2015-07-31
Install string bugs


tavinathanson:
@iskandr found the following bugs in the install strings that get printed in `Genome`/`EnsemblRelease` error messages:

* The shell script install string has underscore-ified variables, while the script itself expects dashes.
* `reference_name` is required by the shell script, but is not included in the install string.
* The shell script errors when run with just a FASTA URL, but install strings with just a FASTA URL are suggested to the user.

tavinathanson:
Fixed by #108

PR #108, opened 2015-08-01
Fixes for various issues after the arbitrary GTF PR

tavinathanson:
Addresses:

* https://github.com/hammerlab/pyensembl/issues/107
* https://github.com/hammerlab/pyensembl/issues/106 (with a bunch of related tests)
* https://github.com/hammerlab/pyensembl/issues/105
* https://github.com/hammerlab/pyensembl/issues/104
* https://github.com/hammerlab/pyensembl/issues/103

@iskandr: as for the `varcode.test.test_variants.test_serialization` issue that I mentioned, that appears to just be a manifestation of not downloading Ensembl 77 before running that test. I have a new varcode `mouse` branch commit that makes sure a few more things are auto-downloaded/installed: https://github.com/hammerlab/varcode/commit/68f98b3c6d0fd905ee4e6406f48cecfb0dbe85f4

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/108)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ except ImportError as e:
 if __name__ == '__main__':
     setup(
         name='pyensembl',
-        version="0.6.11",
+        version="0.7.0",
         description="Python interface to ensembl reference genome metadata",
         author="Alex Rubinsteyn",
         author_email="alex {dot} rubinsteyn {at} mssm {dot} edu",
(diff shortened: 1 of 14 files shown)


setup.py
```diff
@@ -40,7 +40,7 @@
 if __name__ == '__main__':
     setup(
         name='pyensembl',
-        version="0.6.11",
+        version="0.7.0",
```
tavinathanson: Bumping to 0.7 because of the pickling change; thoughts?

iskandr: Seems good, for pickling and other changes (since the Genome API might be a surprising difference as well)

iskandr:
:+1:

PR #114, opened 2015-08-14
Refactoring to support species in EnsemblRelease and to simplify download/install logic

iskandr:
* Added `annotation_name` and `annotation_version` properties to `Genome`, which are used (along with `reference_name`) to create different directories for each genome annotation database.
* Renamed `compute_cache` to `memory_cache`, require objects that use it to explicitly create a `MemoryCache` instance (rather than relying on global state). 
* Hiding the ugliness of direct interactions with `datacache` behind a `DownloadCache` class
* * Got rid of `GenomeSource`: if a file needs to be downloaded, then it's down directly by `Genome` via its `download_cache` instance. 
* All wrappers around data (GTF, SequenceData, Database objects) expect their data to already be locally present (and thus can just take paths instead of `GenomeSource` objects)
* Moved `install_string` onto Genome, got rid of `install_string_python`.
* Added `Species` objects, which track names, synonyms, and the Ensembl releases associated with particular reference assemblies for each species.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/114)


tavinathanson:
Not done yet, but some comments to get the review started

---

Reviewed 16 of 36 files at r1.
Review status: 16 of 36 files reviewed at latest revision, 15 unresolved discussions, all commit checks successful.

---

<sup>**[pyensembl/download_cache.py, line 24 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1CU-r_COj8cXVvgVa)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L24)):</sup>
Shouldn't these all default to None, given your None checks below?

---

<sup>**[pyensembl/ensembl_url_templates.py, line 27 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1LjWWMakGdzf9iA9h)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/ensembl_url_templates.py#L27)):</sup>
Excellent!

---

<sup>**[pyensembl/genome.py, line 181 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1D3O3C-GXG0bRp6wj)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/genome.py#L181)):</sup>
This TODO isn't clear to me? Why?

---

<sup>**[pyensembl/shell.py, line 111 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1J_6MdOzdO2FCqfss)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/shell.py#L111)):</sup>
This doesn't describe `delete-sequence-cache`

---

<sup>**[pyensembl/shell.py, line 145 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1Jve64C8HnL1y7XI9)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/shell.py#L145)):</sup>
So the `MAX_ENSEMBL_RELEASE` logic is gone? Then this is no longer valid:

```
To install the latest Ensembl release (default species is human):
    %(prog)s install
```

I'd either remove that docstring text or add add `MAX_ENSEMBL_RELEASE` back in. If the latter, this check here should probably be an assertion (since it wouldn't be hit normally?).

---

<sup>**[pyensembl/shell.py, line 153 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1KwRN8jyP_1zvazOG)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/shell.py#L153)):</sup>
`delete` vs. `delete-sequence-cache` is pretty confusing to me. If there's an option to delete everything and an option to delete just transcripts, why isn't there also an option to delete just the annotation data, too? Can we just have `delete` and not go more granular?

---

<sup>**[pyensembl/species.py, line 41 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1MoTan7fnHUMppbde)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/species.py#L41)):</sup>
Add a check to make sure that the reference assembly ranges don't overlap?

---

<sup>**[pyensembl/species.py, line 90 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1NZjjo7YprOauczGY)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/species.py#L90)):</sup>
Why is this normalization necessary vs. just requiring that `add_species` be given a proper Ensembl species name (e.g. `homo_sapiens`) that'll work with the URL template? After all, it doesn't seem like `add_species` is exposed elsewhere?

---

<sup>**[pyensembl/species.py, line 128 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1ON6Midu_ZnB0Quoo)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/species.py#L128)):</sup>
Being in the business of keeping all species up to date seems undesirable. We can't just use http://useast.ensembl.org/info/website/archives/assembly.html?redirect=no or something? I think you were asking around about a data source for this?

---

<sup>**[pyensembl/species.py, line 135 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1NyrgRIEh78g2RpVJ)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/species.py#L135)):</sup>
Are synonyms necessary? I just worry about scope screep of this file. If you're concerned with mouse genome data, don't you know what `mus_musculus` means? I don't feel strongly one way or the other; just thinking out loud.

---

<sup>**[pyensembl/species.py, line 136 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1M9EXqfBCqfrnrE2A)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/species.py#L136)):</sup>
Not needed in this PR?

---

<sup>**[pyensembl/transcript.py, line 388 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1Ocsj9kPNBmkN7V9u)** ([raw file](https://github.com/hammerlab/pyensembl/blob/d9e4f1cf44667d876d0513b2d1a937fd11675f95/pyensembl/transcript.py#L388)):</sup>
Why remove this? The resultant error seems like it'd be pretty uninformative?

---

<sup>**[test/test_missing_genome_sources.py, line 55 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1PeDBkTNMeaNxC5ZC)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/test/test_missing_genome_sources.py#L55)):</sup>
Unclear why you went from `_test_mouse_subset` above to `_test_mouse_ensembl81_subset` here

---

<sup>**[test/test_mouse.py, line 34 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1Pk68Iqc1avDPYFFe)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/test/test_mouse.py#L34)):</sup>
Stick with `_test_mouse_ensembl81_subset`?

---

<sup>**[test/test_mouse.py, line 38 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1Pr7viTUk3OF9bc7C)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/test/test_mouse.py#L38)):</sup>
Cleaner/more idiomatic to have a test teardown method that clears the `cache`?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pyensembl/114)

tavinathanson:
Cool PR! I really like how you simplified `GTF` and `SequenceData`. I do have a bunch of comments/questions, though :)

---

Reviewed 20 of 36 files at r1.
Review status: all files reviewed at latest revision, 49 unresolved discussions, all commit checks successful.

---

<sup>**[pyensembl/database.py, line 70 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5TzS-Qo-OsSFuJbAe)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/database.py#L70)):</sup>
Not sure I agree that a GTF is equal to another GTF, even if it has a different cache_subdirectory? (Or a different `create_database_if_missing`, though that's a bit of a weaker argument?)

---

<sup>**[pyensembl/database.py, line 260 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5UMV2jYGV3BGeA0WQ)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/database.py#L260)):</sup>
Don't we want to tell them how to index it?

---

<sup>**[pyensembl/download_cache.py, line 53 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5NwPRn2ck9Y7bXjYJ)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L53)):</sup>
Overall, `DownloadCache` looks nice. Feels a little bloated, but hopefully less so after some of these comments are addressed.

---

<sup>**[pyensembl/download_cache.py, line 70 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5IYIUY1uhplxqWYMG)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L70)):</sup>
This isn't an actual argument to the constructor?

---

<sup>**[pyensembl/download_cache.py, line 79 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5IbVo3d17wDc3lH3S)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L79)):</sup>
Mark as optional?

---

<sup>**[pyensembl/download_cache.py, line 86 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5KPgMlhRo5KPTtlMW)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L86)):</sup>
Do we really need to provide this option vs. always doing this? More options --> more bugs.

---

<sup>**[pyensembl/download_cache.py, line 90 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5KfW56uGqEGFtBsLj)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L90)):</sup>
This isn't a thing?

---

<sup>**[pyensembl/download_cache.py, line 95 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5JNlY4ToBYF2BEykN)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L95)):</sup>
Clarify that the key is the name of the source and the value is the URL? Also, clarify what happens when none is provided?

---

<sup>**[pyensembl/download_cache.py, line 200 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5K1LR-2qKvmY2jgJa)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L200)):</sup>
My preference would be to break this up into `_download_if_necesssary` and `_copy_if_necessary` for readability.

---

<sup>**[pyensembl/download_cache.py, line 228 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5NM-BE4jkZ7XjmTmC)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L228)):</sup>
Minor, but this `ValueError` copy paste seems wasteful. How about something like:

`raise ValueError("Missing genome data file from %s, run: %s" % (
                missing_urls[0] if len(missing_urls) == 1 else missing_urls, install_string)`

---

<sup>**[pyensembl/download_cache.py, line 246 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5No4duFXSQeZIUnMw)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L246)):</sup>
Why do we catch `MissingRemoteFile` but not `MissingLocalFile`? I assume we should catch that, and maybe worth adding a test that would fail if we didn't?

---

<sup>**[pyensembl/download_cache.py, line 249 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5Ls2lKcjUd0AiJOrN)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L249)):</sup>
This isn't used anywhere? Can we just get rid of it? (I hope we can, because I also don't like having two similar functions for `path` and for `paths`.)

---

<sup>**[pyensembl/download_cache.py, line 280 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5LyMiPyLi1VW2n8Jv)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L280)):</sup>
I think renaming these two to `delete_cached_files` and `delete_all_cached_files` would be more clear.

---

<sup>**[pyensembl/ensembl_release.py, line 42 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5Z2GdoSwCfB-8Q2a0)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/ensembl_release.py#L42)):</sup>
If `species.py` is only applicable to Ensembl, maybe rename to `ensembl_species.py`?

---

<sup>**[pyensembl/ensembl_release.py, line 80 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5J0yO0eKnAMoajyFr)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/ensembl_release.py#L80)):</sup>
Nit: you call this `missing_files` in `Genome` vs. `missing_fields`

---

<sup>**[pyensembl/ensembl_release.py, line 81 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5ZZoO4wxM6Z6y5jMP)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/ensembl_release.py#L81)):</sup>
I'm still confused about whether installing is even possible in Python anymore; but if it is, why not display how to do it in Python? I guess I don't feel too strongly.

---

<sup>**[pyensembl/ensembl_release.py, line 90 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5_6vg5VhTF8MszeNN)** ([raw file](https://github.com/hammerlab/pyensembl/blob/d9e4f1cf44667d876d0513b2d1a937fd11675f95/pyensembl/ensembl_release.py#L90)):</sup>
Where is the logic now for adding the release to the FASTA file? I assume it's somewhere and I just missed it?

---

<sup>**[pyensembl/genome.py, line 184 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5ShQsMtNnx9QIG2Z-)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/genome.py#L184)):</sup>
Why isn't this called `install` if it's called when the `shell.py` argument == `install`? Not arguing about names; just about consistency.

---

<sup>**[pyensembl/genome.py, line 212 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5a8pxz7jpSvk-qszH)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/genome.py#L212)):</sup>
Didn't we talk about moving away from some of the lazy loading stuff; was that totally separate from this?

---

<sup>**[pyensembl/genome.py, line 231 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5OrOki0nIG_9cR6Iu)** ([raw file](https://github.com/hammerlab/pyensembl/blob/d9e4f1cf44667d876d0513b2d1a937fd11675f95/pyensembl/genome.py#L231)):</sup>
What does installing without the shell script look like, now? `genome.load_all_data()`? What happened to the index/download distinction, why the rename, and why don't the tests need to call `load_all_data`?

---

<sup>**[pyensembl/genome.py, line 239 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5OR3RQjpTZ_F_1TDj)** ([raw file](https://github.com/hammerlab/pyensembl/blob/d9e4f1cf44667d876d0513b2d1a937fd11675f95/pyensembl/genome.py#L239)):</sup>
Why remove `index`? I find that pretty useful for re-creating the DB without re-downloading the data. I thought you did too?

---

<sup>**[pyensembl/genome.py, line 280 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5L6R0e6lZyW327B-j)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/genome.py#L280)):</sup>
So clearing the cache leaves GTF and FASTA, but calling `delete` deletes everything? This is non-obvious to me and I don't think would be obvious to others. Can you say more about your thought process for when/where certain file types should be deleted?

---

<sup>**[pyensembl/gtf.py, line 40 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5PJ6_GfpT_hOYpPQB)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/gtf.py#L40)):</sup>
I feel like we're taking a small step backwards, here. (Though clearly, removing all the gunk in `GTF` and `SequenceData` is a large step forward.) We talked about moving this path splitting/handling logic outside of `GTF` etc., and now it's back?

---

<sup>**[pyensembl/gtf.py, line 66 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5UmLzX1VoLdg-aYBd)** ([raw file](https://github.com/hammerlab/pyensembl/blob/d9e4f1cf44667d876d0513b2d1a937fd11675f95/pyensembl/gtf.py#L66)):</sup>
Super awesome that you were able to remove all this gunk!

---

<sup>**[pyensembl/sequence_data.py, line 32 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5QfowjWA37FTrSvoN)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/sequence_data.py#L32)):</sup>
So we went from a `fasta_source` to needing `fasta_path`, `sequence_type` and `cache_directory_path`, and needing to do path munging inside `SequenceData`. That again seems like a small step backwards to me, like with `GTF`.

---

<sup>**[pyensembl/sequence_data.py, line 87 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5UrO28_72A3EPlhyD)** ([raw file](https://github.com/hammerlab/pyensembl/blob/d9e4f1cf44667d876d0513b2d1a937fd11675f95/pyensembl/sequence_data.py#L87)):</sup>
Sayonara, download/cache gunk! Woo!

---

<sup>**[pyensembl/sequence_data.py, line 90 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5RgBHt4Eb9TqP4dWe)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/sequence_data.py#L90)):</sup>
Woohoo, moving away from BioPython!

---

<sup>**[pyensembl/transcript.py, line 465 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5T7s8g2u2p1A0o3zI)** ([raw file](https://github.com/hammerlab/pyensembl/blob/d9e4f1cf44667d876d0513b2d1a937fd11675f95/pyensembl/transcript.py#L465)):</sup>
And why remove this?

---

<sup>**[requirements.txt, line 4 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5Rrrg9N2s56uqQ79q)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/requirements.txt#L4)):</sup>
Do we still need biopython?

---

<sup>**[setup.py, line 68 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5RtOSRdTK34hPEs76)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/setup.py#L68)):</sup>
(Same question about needing biopython here?)

---

<sup>**[test/test_download.py, line 0 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5S39pRTiuUDKu6qbF)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/test/test_download.py#L0)):</sup>
The logic to "Return True if the GTF database gets created, which should be different depending on whether the database already existed." seems like it's still applicable and still worth testing?

---

<sup>**[test/test_download_cache.py, line 10 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5SDOB3YTjb-HRffNk)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/test/test_download_cache.py#L10)):</sup>
This doesn't read as clearly as using `assert_raises`

Also, probably worth at least adding a test for remote files as well?

---

<sup>**[test/test_sequence_data.py, line 34 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5TOnvc42RN9Ei60pu)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/test/test_sequence_data.py#L34)):</sup>
In general, checking `ValueError` seems problematic since it could be any `ValueError`. I suggest either changing to a custom error or checking the `ValueError` text.

---

<sup>**[test/test_ucsc_gtf.py, line 60 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5TdntweCBjnQcibyd)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/test/test_ucsc_gtf.py#L60)):</sup>
So why do we not need to call `genome.load_all_data()`?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pyensembl/114)

iskandr:
Review status: all files reviewed at latest revision, 33 unresolved discussions, all commit checks successful.

---

<sup>**[pyensembl/database.py, line 70 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5TzS-Qo-OsSFuJbAe)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/database.py#L70)):</sup>
I'm thinking of the GTF object as an in-memory representation of a GTF file. Why wouldn't they be equal to each other? 

---

<sup>**[pyensembl/database.py, line 260 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5UMV2jYGV3BGeA0WQ)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/database.py#L260)):</sup>
I think that instead I'm just shifting the responsibility for making sure stuff exists to `Genome` and simplifying the amount of logic allowed in `Database`. 

That fits with the general trend of this PR of centralizing anything tricky and having the "leaf" objects be dumber and do less. 

---

<sup>**[pyensembl/download_cache.py, line 70 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5IYIUY1uhplxqWYMG)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L70)):</sup>
It was at one point :-)

---

<sup>**[pyensembl/download_cache.py, line 86 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5KPgMlhRo5KPTtlMW)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L86)):</sup>
I think it's actually better not to do this by default (since it eliminates the need to "install" fully local genomes), but yeah we can just always stick with one behavior. 

---

<sup>**[pyensembl/download_cache.py, line 246 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5No4duFXSQeZIUnMw)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L246)):</sup>
`MissingLocalFile` is a legit exception that the user needs to take care of (i.e. the file they think exists locally doesn't)

---

<sup>**[pyensembl/download_cache.py, line 249 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5Ls2lKcjUd0AiJOrN)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L249)):</sup>
Yeah, this is the old version hanging around, got rid of it. 

---

<sup>**[pyensembl/download_cache.py, line 280 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5LyMiPyLi1VW2n8Jv)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L280)):</sup>
I changed them to delete_cached_files and delete_cache_directory

---

<sup>**[pyensembl/ensembl_release.py, line 42 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5Z2GdoSwCfB-8Q2a0)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/ensembl_release.py#L42)):</sup>
Maybe whenever we have more direct support for RefSeq? The current "generality" of PyEnsembl still seems preparatory/aspirational. 

---

<sup>**[pyensembl/ensembl_release.py, line 80 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5J0yO0eKnAMoajyFr)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/ensembl_release.py#L80)):</sup>
Changed to `missing_urls_dict`. 

---

<sup>**[pyensembl/ensembl_release.py, line 81 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5ZZoO4wxM6Z6y5jMP)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/ensembl_release.py#L81)):</sup>
It's still possible but since these insides are shifting, let's not expose it to the user and instead give them only One Way To Do It. 

---

<sup>**[pyensembl/ensembl_release.py, line 90 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5_6vg5VhTF8MszeNN)** ([raw file](https://github.com/hammerlab/pyensembl/blob/d9e4f1cf44667d876d0513b2d1a937fd11675f95/pyensembl/ensembl_release.py#L90)):</sup>
We don't need it anymore, since releases are isolated within their own directories. 

---

<sup>**[pyensembl/genome.py, line 181 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1D3O3C-GXG0bRp6wj)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/genome.py#L181)):</sup>
Any use of subdirectories is because we're relying on datacache to figure out the global path for us, which is something we need to move away from to then cut the dependency on datacache (or make datacache smaller and more focused). 

---

<sup>**[pyensembl/genome.py, line 184 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5ShQsMtNnx9QIG2Z-)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/genome.py#L184)):</sup>
Because it loads all the data into memory *and* if auto_download is True then it installs whatever is missing. That `install` from the shell is implemented using this function is an implementation detail. 

---

<sup>**[pyensembl/genome.py, line 212 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5a8pxz7jpSvk-qszH)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/genome.py#L212)):</sup>
I tried for a while to do all the data loading up front but couldn't find an easy way to make it work with constructing `EnsemblRelease` objects but not immediately downloading/indexing their data (which you want to do to e.g. clear their cache directory). 

---

<sup>**[pyensembl/genome.py, line 231 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5OrOki0nIG_9cR6Iu)** ([raw file](https://github.com/hammerlab/pyensembl/blob/d9e4f1cf44667d876d0513b2d1a937fd11675f95/pyensembl/genome.py#L231)):</sup>
`Genome(..., auto_download=True).load_all_data()`. But, there shouldn't be much reason to do this outside the shell script. 

I don't think there's any reason to expose indexing to the user since all I ever find myself wanting to do with index files is *delete* them because they're outdated in some way. 

The tests don't need to call `load_all_data` because they either depend on having a large number of Ensembl releases installed or use local files which get used in-place.

---

<sup>**[pyensembl/genome.py, line 239 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5OR3RQjpTZ_F_1TDj)** ([raw file](https://github.com/hammerlab/pyensembl/blob/d9e4f1cf44667d876d0513b2d1a937fd11675f95/pyensembl/genome.py#L239)):</sup>
I do, you can achieve the same effect with 

```
pyensembl delete-sequence-cache --release 75
pyensembl install --release 75 
```

We can add a command to e.g. delete-cache or delete-gtf-cache whenever we need those cases. 

---

<sup>**[pyensembl/genome.py, line 280 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5L6R0e6lZyW327B-j)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/genome.py#L280)):</sup>
This is really a hack, I'll more comments explaining what's going on. But, we download fasta and gtf files (and I guess their gzipped versions) and only create pickle and db files as temporaries. I switched this code to at least delete the expected "temporary" filetypes rather than delete everything other than FASTA/GTF files. 

---

<sup>**[pyensembl/gtf.py, line 40 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5PJ6_GfpT_hOYpPQB)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/gtf.py#L40)):</sup>
Yeah, I wasn't totally thrilled with having path munging this deep into the object graph but I don't know how else to co-locate the GTF temporary files with the other cached data. That said, the way I'm saving .CSV snapshots of the GTF is questionably useful, since I rarely look at GTF subsets as a dataframe (the way I originally envisioned I would want to). So maybe we can eventually get rid of this? Particular if we can speed up GTF parsing so that reading a GTF file twice wouldn't be devastatingly slow. 

---

<sup>**[pyensembl/sequence_data.py, line 32 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5QfowjWA37FTrSvoN)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/sequence_data.py#L32)):</sup>
On the other hand, these objects are more independent and testable. A SequenceData object taking a FASTA path directly and having a free parameter for where you tell it to store derived/cached files seems fine to me. 

---

<sup>**[pyensembl/shell.py, line 145 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1Jve64C8HnL1y7XI9)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/shell.py#L145)):</sup>
I'm removing the default docstring since I'd rather have the user do stuff more explicitly. 

---

<sup>**[pyensembl/shell.py, line 153 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1KwRN8jyP_1zvazOG)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/shell.py#L153)):</sup>
Just `delete` also nukes the (large) downloaded GTF and FASTA files. I'd rather add more fine-grained delete options as we need them. 

---

<sup>**[pyensembl/species.py, line 128 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1ON6Midu_ZnB0Quoo)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/species.py#L128)):</sup>
I'm not happy about this either but it's the only easy way to get mice working for now. I wanted to get the species representation wired together first, maybe we can do a later PR to auto-populate from remote data?

---

<sup>**[pyensembl/species.py, line 135 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1NyrgRIEh78g2RpVJ)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/species.py#L135)):</sup>
I'd rather species = "human" or "mouse" than "mus_musculus". 

---

<sup>**[pyensembl/species.py, line 136 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1M9EXqfBCqfrnrE2A)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/species.py#L136)):</sup>
?

---

<sup>**[pyensembl/transcript.py, line 388 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1Ocsj9kPNBmkN7V9u)** ([raw file](https://github.com/hammerlab/pyensembl/blob/d9e4f1cf44667d876d0513b2d1a937fd11675f95/pyensembl/transcript.py#L388)):</sup>
`transcript_sequences` is a property which can raise its own errors (and thus doesn't need to be checked externally)

---

<sup>**[pyensembl/transcript.py, line 465 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5T7s8g2u2p1A0o3zI)** ([raw file](https://github.com/hammerlab/pyensembl/blob/d9e4f1cf44667d876d0513b2d1a937fd11675f95/pyensembl/transcript.py#L465)):</sup>
^ same as above

---

<sup>**[setup.py, line 68 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5RtOSRdTK34hPEs76)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/setup.py#L68)):</sup>
Nope, removing

---

<sup>**[test/test_download.py, line 0 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5S39pRTiuUDKu6qbF)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/test/test_download.py#L0)):</sup>
These mocked out tests can be tedious since they are tied to the internals in a way that's annoying to update when the internals change. I was thinking about fixing this test but we're still due for more changes down the road and I don't think I have the emotional energy to maintain tests that know about internal state. 

---

<sup>**[test/test_missing_genome_sources.py, line 55 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1PeDBkTNMeaNxC5ZC)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/test/test_missing_genome_sources.py#L55)):</sup>
Just a missing copy/paste, the directory name doesn't really matter though.

---

<sup>**[test/test_mouse.py, line 34 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1Pk68Iqc1avDPYFFe)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/test/test_mouse.py#L34)):</sup>
Sure, but I don't think it actually matters which directories we use for small test data. We'll end up with some new tests that maybe use a different directory and I think that's probably fine.

---

<sup>**[test/test_sequence_data.py, line 34 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5TOnvc42RN9Ei60pu)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/test/test_sequence_data.py#L34)):</sup>
Checking ValueError text seems really future-unfriendly. I'll just move the raises check further in. 

---

<sup>**[test/test_ucsc_gtf.py, line 60 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5TdntweCBjnQcibyd)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/test/test_ucsc_gtf.py#L60)):</sup>
(1) It's a local path, so the files don't need to be copied to the cache

(2) `Genome.genes()` uses the `Genome.db` property which triggers lazy loading of the GTF and construction of the associated database. 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pyensembl/114)

iskandr:
Review status: 24 of 36 files reviewed at latest revision, 33 unresolved discussions, all commit checks successful.

---

<sup>**[pyensembl/species.py, line 90 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1NZjjo7YprOauczGY)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/species.py#L90)):</sup>
normalize_species_name gets used for user-supplied names also, I guess it's not necessary to call it internally. 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pyensembl/114)

tavinathanson:
Offline discussion: Alex clarified that this PR moves from install/index/download as useful verbs to simply installing and deleting. So installing handles both downloading and indexing.

Currently, there's no `install` verb in Python anymore, and I think that's not so great. Whether it's named `install` or something else, I think the user should never be surprised by a lengthy indexing process. In the remote case, this would be more rare because the user needs to download first; and when they download, they'll be installing, so indexing will happen too. But in the local case, `install` is pretty helpful so that the user isn't surprising by indexing when they just tried to get a gene.

If that's not convincing (because maybe we don't care about the local case that much yet): there's also `auto_download`. If we're moving away from `download` as a verb, I think we should move away from `auto_download` (as I'm already getting confused by its existence in this new world). How about renaming it to `auto_install`, and then keeping `install` required unless the user `auto_install`s? Then we're more fully moved away from `download`/`index`, and we tackle the local file problem too.

---

Reviewed 13 of 13 files at r2.
Review status: all files reviewed at latest revision, 24 unresolved discussions, all commit checks successful.

---

<sup>**[pyensembl/database.py, line 70 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5TzS-Qo-OsSFuJbAe)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/database.py#L70)):</sup>
Looks like you changed `__eq__`; shouldn't `hash` match that?

---

<sup>**[pyensembl/download_cache.py, line 86 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5KPgMlhRo5KPTtlMW)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L86)):</sup>
Like I mentioned in another comment, I think `install` is a useful verb for local genomes. The user still needs to index, and `install` covers that.

---

<sup>**[pyensembl/download_cache.py, line 246 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5No4duFXSQeZIUnMw)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L246)):</sup>
Okay, but isn't `MissingRemoteFile` also a legit exception? But with that, you're calling `raise_missing_file_error` with `field_name` and `path_or_url` for a more complete error message; but with `MissingLocalFile` I don't believe you are?

---

<sup>**[pyensembl/ensembl_release.py, line 81 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5ZZoO4wxM6Z6y5jMP)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/ensembl_release.py#L81)):</sup>
Okay, that's a fair argument. I like simplicity!

---

<sup>**[pyensembl/ensembl_release.py, line 90 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5_6vg5VhTF8MszeNN)** ([raw file](https://github.com/hammerlab/pyensembl/blob/d9e4f1cf44667d876d0513b2d1a937fd11675f95/pyensembl/ensembl_release.py#L90)):</sup>
Ah, right. Sweet. That does sound much cleaner :)

---

<sup>**[pyensembl/genome.py, line 181 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1D3O3C-GXG0bRp6wj)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/genome.py#L181)):</sup>
Maybe say something like that in the comment instead of what you have there?

---

<sup>**[pyensembl/genome.py, line 184 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5ShQsMtNnx9QIG2Z-)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/genome.py#L184)):</sup>
My current understanding of `install` is indexing and downloading. I'm actually confused about what `auto_download` does now. Should it be called `auto_install` and simply result in not needing to `install()`?

---

<sup>**[pyensembl/genome.py, line 212 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5a8pxz7jpSvk-qszH)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/genome.py#L212)):</sup>
Isn't that the purpose of having an `install` method? I understand moving away from `index` and `download`, but why not require `install`? Wouldn't that also eliminate the local-file-results-in-a-10-minute-wait issue, since we'd just require them to call `install`?

Even for local files, `install` still makes sense to me, since we need the DB and transcript indexing?

---

<sup>**[pyensembl/genome.py, line 231 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5OrOki0nIG_9cR6Iu)** ([raw file](https://github.com/hammerlab/pyensembl/blob/d9e4f1cf44667d876d0513b2d1a937fd11675f95/pyensembl/genome.py#L231)):</sup>
I'm on board with the removal of `index` and `download`, and I have other comments addressing the other points.

---

<sup>**[pyensembl/genome.py, line 239 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5OR3RQjpTZ_F_1TDj)** ([raw file](https://github.com/hammerlab/pyensembl/blob/d9e4f1cf44667d876d0513b2d1a937fd11675f95/pyensembl/genome.py#L239)):</sup>
Except, like we talked about, I do use the GTF case a lot already :)

How about:

`pyensembl delete --release 75`
`pyensembl delete --release 75 --type fasta`
`pyensembl delete --release 75 --type gtf`

Less verbs?

---

<sup>**[pyensembl/genome.py, line 280 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5L6R0e6lZyW327B-j)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/genome.py#L280)):</sup>
I'm still confused, even after our chat, about how this relates to `delete-sequence-data`. Do you typically need to delete the `FASTA` files more or the `pickle` files more? For me, I typically need to delete the `.db` file more than the `GTF` file. I would expect `delete-sequence-data` to delete both the `FASTA` and `pickle` files.

---

<sup>**[pyensembl/genome.py, line 286 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx79bCgYsiD_jft_HBr)** ([raw file](https://github.com/hammerlab/pyensembl/blob/53f037cf9e7f8d60007ce67a001decbb0867d35e/pyensembl/genome.py#L286)):</sup>
Doesn't SequenceData know about `.pickle`? And doesn't GTF know about `.db`? Can't we use their cache clearing mechanisms?

---

<sup>**[pyensembl/gtf.py, line 40 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5PJ6_GfpT_hOYpPQB)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/gtf.py#L40)):</sup>
I think the fact that `SequenceData` and `GTF` basically repeat the path munging is a sign that the code can and should be shared between them. I'll let this go for now, though :)

---

<sup>**[pyensembl/shell.py, line 145 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1Jve64C8HnL1y7XI9)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/shell.py#L145)):</sup>
That sounds good to me!

---

<sup>**[pyensembl/shell.py, line 153 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1KwRN8jyP_1zvazOG)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/shell.py#L153)):</sup>
I completely agree with not adding options before we need them, but this current situation still feels confusing. Do we really need `delete-sequence-cache`? I can't imagine many users thinking about the distinction properly, which makes it fairly unhelpful.

---

<sup>**[pyensembl/species.py, line 90 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1NZjjo7YprOauczGY)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/species.py#L90)):</sup>
So we're trusting that a user is going to supply a name that magically will fit into an Ensembl URL when lowercased and underscorized? That seems flawed, if I'm understanding correctly?

---

<sup>**[pyensembl/species.py, line 128 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1ON6Midu_ZnB0Quoo)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/species.py#L128)):</sup>
Cool, I'll file an issue.

---

<sup>**[pyensembl/species.py, line 135 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1NyrgRIEh78g2RpVJ)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/species.py#L135)):</sup>
I guess it is easier to type. Fair enough.

---

<sup>**[pyensembl/species.py, line 136 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1M9EXqfBCqfrnrE2A)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/species.py#L136)):</sup>
I'm unclear what this TODO is saying, and whether that (fixing release range; whatever it means) should have happened in this PR.

---

<sup>**[test/test_download.py, line 0 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5S39pRTiuUDKu6qbF)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/test/test_download.py#L0)):</sup>
I think it really depends on how important the thing we're testing is. In this case, I concede that this test is helpful but not critical. (But if it were more critical, I'd be arguing with you.)

---

<sup>**[test/test_missing_genome_sources.py, line 55 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1PeDBkTNMeaNxC5ZC)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/test/test_missing_genome_sources.py#L55)):</sup>
Yeah, was just a nit consistency thing.

---

<sup>**[test/test_sequence_data.py, line 34 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5TOnvc42RN9Ei60pu)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/test/test_sequence_data.py#L34)):</sup>
I agree that message-checking is pretty ugly, but I also stand by checking for `ValueError` being unhelpful. Moving it further seems like a reasonable compromise for now.

---

<sup>**[test/test_ucsc_gtf.py, line 60 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5TdntweCBjnQcibyd)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/test/test_ucsc_gtf.py#L60)):</sup>
But DB indexing can take quite a while. Just like the `auto_downloading` discussion from way back when, I feel like we don't want to surprise users with a 10 minute delay when they're just grabbing a gene count.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pyensembl/114)

iskandr:
I got rid of auto_install. If a user creates a Genome using local files they now have to run `genome.index` for things to work correctly. If they don't run it, they'll get an error message with a commandline install string that will run index for them.

---

Review status: 25 of 36 files reviewed at latest revision, 7 unresolved discussions, all commit checks successful.

---

<sup>**[pyensembl/download_cache.py, line 253 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5No4duFXSQeZIUnMw)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L253)):</sup>
Add an issue for this test?

---

<sup>**[pyensembl/genome.py, line 239 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5OR3RQjpTZ_F_1TDj)** ([raw file](https://github.com/hammerlab/pyensembl/blob/d9e4f1cf44667d876d0513b2d1a937fd11675f95/pyensembl/genome.py#L239)):</sup>
Seems way easier to delete the wrong thing, I switched it to --delete-all-files and --delete-index-files. 

---

<sup>**[pyensembl/genome.py, line 286 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx79bCgYsiD_jft_HBr)** ([raw file](https://github.com/hammerlab/pyensembl/blob/53f037cf9e7f8d60007ce67a001decbb0867d35e/pyensembl/genome.py#L286)):</sup>
Switched to calling clear_cache() and having a method which explicitly deletes the Database's created db file

---

<sup>**[pyensembl/gtf.py, line 40 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5PJ6_GfpT_hOYpPQB)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/gtf.py#L40)):</sup>
Make an issue for it?

---

<sup>**[pyensembl/species.py, line 92 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1NZjjo7YprOauczGY)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/species.py#L92)):</sup>
You still need the latin name to actually be in the species dictionary, so we're not just feeding any user input to a URL. 

---

<sup>**[pyensembl/species.py, line 150 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx1M9EXqfBCqfrnrE2A)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/species.py#L150)):</sup>
The range given for mouse releases isn't right and we don't have ranges for other assemblies.  (just copied that into the file)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pyensembl/114)

tavinathanson:
Reviewed 2 of 11 files at r3.
Review status: 27 of 36 files reviewed at latest revision, 14 unresolved discussions, all commit checks successful.

---

<sup>**[pyensembl/__init__.py, line 49 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-JxFDb3J8UjLNmtOIASX)** ([raw file](https://github.com/hammerlab/pyensembl/blob/53f037cf9e7f8d60007ce67a001decbb0867d35e/pyensembl/__init__.py#L49)):</sup>
What about the tests that use auto_download? Do all unit tests run?

---

<sup>**[pyensembl/database.py, line 56 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-JxFDvdlXEIu9xXs8kWe)** ([raw file](https://github.com/hammerlab/pyensembl/blob/dde05babc2b98eae59ccf11ddd035208f099c54d/pyensembl/database.py#L56)):</sup>
Feels weird to just pass in the string here, but we can refactor it later.

---

<sup>**[pyensembl/download_cache.py, line 208 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-JxFD-ORoGXRx8bYxqsm)** ([raw file](https://github.com/hammerlab/pyensembl/blob/dde05babc2b98eae59ccf11ddd035208f099c54d/pyensembl/download_cache.py#L208)):</sup>
This comment is no longer valid

---

<sup>**[pyensembl/genome.py, line 128 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-JxFDJtxyt8QsxiNCOUG)** ([raw file](https://github.com/hammerlab/pyensembl/blob/dde05babc2b98eae59ccf11ddd035208f099c54d/pyensembl/genome.py#L128)):</sup>
This feels like a weird way of removing `auto-download`. Now auto download is still there, but with a different name? I'm fine to remove `auto-download` if we actually remove it, but just remove it at the root and renaming it everywhere else doesn't seem complete. Thoughts?

---

<sup>**[pyensembl/genome.py, line 178 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-JxFEu-aXRJNzZQDh56_)** ([raw file](https://github.com/hammerlab/pyensembl/blob/dde05babc2b98eae59ccf11ddd035208f099c54d/pyensembl/genome.py#L178)):</sup>
Wait, so we're back to `auto_download`ing without the word `auto_download`? If a URL is provided, we download it instead of erroring? I'm confused :\

---

<sup>**[pyensembl/shell.py, line 146 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-JxFEUrcBa32Y_f70OQl)** ([raw file](https://github.com/hammerlab/pyensembl/blob/dde05babc2b98eae59ccf11ddd035208f099c54d/pyensembl/shell.py#L146)):</sup>
I really don't understand why we our shell script verbs are different from our Python verbs?

---

<sup>**[test/test_missing_genome_sources.py, line 29 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-JxFEbJi9zVl9EXCfhtZ)** ([raw file](https://github.com/hammerlab/pyensembl/blob/dde05babc2b98eae59ccf11ddd035208f099c54d/test/test_missing_genome_sources.py#L29)):</sup>
When does downloading happen? And why are we propagating the idea of `indexing` that's separate from downloading even more than we were before?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pyensembl/114)

iskandr:
Review status: 27 of 36 files reviewed at latest revision, 14 unresolved discussions, all commit checks successful.

---

<sup>**[pyensembl/__init__.py, line 45 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-JxFDb3J8UjLNmtOIASX)** ([raw file](https://github.com/hammerlab/pyensembl/blob/53f037cf9e7f8d60007ce67a001decbb0867d35e/pyensembl/__init__.py#L45)):</sup>
All the unit tests run, are there any still use auto_download? When I run `grep -R auto_download *` I don't see anything in the tests directory.

---

<sup>**[pyensembl/database.py, line 56 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-JxFDvdlXEIu9xXs8kWe)** ([raw file](https://github.com/hammerlab/pyensembl/blob/dde05babc2b98eae59ccf11ddd035208f099c54d/pyensembl/database.py#L56)):</sup>
Yeah I agree, but I wasn't sure how to handle the "not yet created" case if Database isn't always auto-creating. 

---

<sup>**[pyensembl/download_cache.py, line 208 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-JxFD-ORoGXRx8bYxqsm)** ([raw file](https://github.com/hammerlab/pyensembl/blob/dde05babc2b98eae59ccf11ddd035208f099c54d/pyensembl/download_cache.py#L208)):</sup>
I'm confused, which part is no longer valid? 

---

<sup>**[pyensembl/genome.py, line 128 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-JxFDJtxyt8QsxiNCOUG)** ([raw file](https://github.com/hammerlab/pyensembl/blob/dde05babc2b98eae59ccf11ddd035208f099c54d/pyensembl/genome.py#L128)):</sup>
There's no global auto-download associated with the Genome object but the helper functions do have it for use by the Genome.download() method. Does that make sense?

---

<sup>**[pyensembl/genome.py, line 178 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-JxFEu-aXRJNzZQDh56_)** ([raw file](https://github.com/hammerlab/pyensembl/blob/dde05babc2b98eae59ccf11ddd035208f099c54d/pyensembl/genome.py#L178)):</sup>
Oh, that's a copy/paste error. Should be download_if_missing=download_if_missing.

---

<sup>**[pyensembl/shell.py, line 146 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-JxFEUrcBa32Y_f70OQl)** ([raw file](https://github.com/hammerlab/pyensembl/blob/dde05babc2b98eae59ccf11ddd035208f099c54d/pyensembl/shell.py#L146)):</sup>
We have Python methods that implement the logic of the shell script but I want to keep the shell script as the main user-facing interface with changing what files exist on the file system. If a user needs the download/index methods so be it, but I don't want to provide a `install()` method and tell the user about-- particularly since I think the verb "install" has a weird sound when used within a Python session. 

---

<sup>**[test/test_missing_genome_sources.py, line 29 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-JxFEbJi9zVl9EXCfhtZ)** ([raw file](https://github.com/hammerlab/pyensembl/blob/dde05babc2b98eae59ccf11ddd035208f099c54d/test/test_missing_genome_sources.py#L29)):</sup>
There's no downloading since the file is local and the default behavior in download_cache is to leave local files where they are. 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pyensembl/114)

tavinathanson:
Merge away!

---

Reviewed 8 of 11 files at r3, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 5 unresolved discussions, all commit checks successful.

---

<sup>**[pyensembl/download_cache.py, line 208 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-JxFD-ORoGXRx8bYxqsm)** ([raw file](https://github.com/hammerlab/pyensembl/blob/dde05babc2b98eae59ccf11ddd035208f099c54d/pyensembl/download_cache.py#L208)):</sup>
Ah, `self.copy_local_to_cache` does still exist; nevermind.

---

<sup>**[pyensembl/download_cache.py, line 253 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-Jx5No4duFXSQeZIUnMw)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a97a7cc2225e180fe67e46f272cb1fab199163fd/pyensembl/download_cache.py#L253)):</sup>
Offline discussion: we want to provide an install string in the remote case, but it isn't applicable in the local case. So this is fine.

---

<sup>**[pyensembl/genome.py, line 128 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-JxFDJtxyt8QsxiNCOUG)** ([raw file](https://github.com/hammerlab/pyensembl/blob/dde05babc2b98eae59ccf11ddd035208f099c54d/pyensembl/genome.py#L128)):</sup>
Fair enough; a bit confusing but sounds like it works.

---

<sup>**[pyensembl/shell.py, line 146 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-JxFEUrcBa32Y_f70OQl)** ([raw file](https://github.com/hammerlab/pyensembl/blob/dde05babc2b98eae59ccf11ddd035208f099c54d/pyensembl/shell.py#L146)):</sup>
Eh, I don't agree with this but not worth holding up the PR. These verb discussions are controversial!

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pyensembl/114)

iskandr:
:+1:

---

Review status: all files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[pyensembl/shell.py, line 146 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/114#-JxFEUrcBa32Y_f70OQl)** ([raw file](https://github.com/hammerlab/pyensembl/blob/dde05babc2b98eae59ccf11ddd035208f099c54d/pyensembl/shell.py#L146)):</sup>
We can add `install()` back in if you find it useful!

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pyensembl/114)

PR #117, opened 2015-08-21
Update deprecated skbio functionality to newer equivalnets

iskandr:
* `skbio.parse_fasta` is deprecated (and seems to be missing entirely from the latest pip version), using `skbio.read`

* Similarly, using `skbio.DNA` instead of `skbio.DNASequence`

* Got rid of expectation that sequences returned by `SequenceData` should have `complement()` method (still undecided whether these should be `str` or `skbio.Sequence` objects, Varcode wants to concatenate sequences which skbio doesn't support)

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/117)


tavinathanson:
Looks good to me!

PR #119, opened 2015-08-27
Rewrote GTF parsing to use less memory

iskandr:
The Pandas CSV parser is, for unknown reasons, a memory hog under Python 2. Even if we optimize the memory consumed by the data we're storing, we're still risking running above Travis's 3gb limit just during parsing time. I poked around the Pandas source for a while trying to figure out what's going on but eventually gave up and wrote my own stripped down GTF parsing logic in pure python. Surprisingly, this is a little bit faster and keeps both Py2 & Py3 below 2gb of usage during parse time. 

Other changes:
* interning strings to prevent memory explosion due to existence of millions of identical identifiers (e.g. "gene", "1", or even "TP53"). 
* Fixed double-writing of the parsed CSV in `gtf.py`
* Changed some Ensembl77 tests to use latest GRCh38 release instead
* Removed travis install dependency on Ensembl79, since that's not used at all
* Removed our ill-fated attempt at assigning IDs to exons from old GTFs (`reconstruct_exon_id_column`) since that never actually worked.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/119)


tavinathanson:
Reviewed 12 of 12 files at r1.
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

<sup>**[.travis.yml, line 31 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/119#-JxlNQ4o5il-GP6XxFqN)** ([raw file](https://github.com/hammerlab/pyensembl/blob/4e3f9610b159b0056a153b201b142695e9f5e530/.travis.yml#L31)):</sup>
I think you'll need to re-encrypt this for the `pyensembl` repo? You can download the travis commandline utility and run `travis encrypt [password] -r hammerlab/pyensembl`

The password is on https://hammerlab.atlassian.net/wiki/display/MSL/Releasing+a+Project

---

<sup>**[pyensembl/gtf_parsing.py, line 111 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/119#-JxlQnyTmKS7OK0oRwR0)** ([raw file](https://github.com/hammerlab/pyensembl/blob/4e3f9610b159b0056a153b201b142695e9f5e530/pyensembl/gtf_parsing.py#L111)):</sup>
Remove this line?

---

<sup>**[pyensembl/gtf_parsing.py, line 224 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/119#-JxlRGkUXZXMT-UVMlN7)** ([raw file](https://github.com/hammerlab/pyensembl/blob/4e3f9610b159b0056a153b201b142695e9f5e530/pyensembl/gtf_parsing.py#L224)):</sup>
Is this stuff unit tested? If not, I think it should be with some simple GTF input/dataframe output for each of the cases you're conditioning on? Would be a bummer to accidentally drop some rows (for example) and never know, since that could easily be not discovered by larger tests?

---

<sup>**[pyensembl/gtf_parsing.py, line 238 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/119#-JxlRv2vsXMsV8Zc3UQS)** ([raw file](https://github.com/hammerlab/pyensembl/blob/4e3f9610b159b0056a153b201b142695e9f5e530/pyensembl/gtf_parsing.py#L238)):</sup>
Consistency with `print` vs. `logging.info`?

---

<sup>**[pyensembl/gtf_parsing.py, line 241 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/119#-JxlSoAzZlhvO2qCOvu3)** ([raw file](https://github.com/hammerlab/pyensembl/blob/4e3f9610b159b0056a153b201b142695e9f5e530/pyensembl/gtf_parsing.py#L241)):</sup>
I don't see the single-character string special casing; am I mis-reading?

---

<sup>**[pyensembl/gtf_parsing.py, line 359 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/119#-JxlQ-U6B6vaPaKNCsBz)** ([raw file](https://github.com/hammerlab/pyensembl/blob/a2717397ccf27d83379de1b74ecaef08cf46b778/pyensembl/gtf_parsing.py#L359)):</sup>
So what do you mean that this never worked? Also, are there any related tests that you need to remove now?

---

<sup>**[pyensembl/memory_cache.py, line 67 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/119#-JxlPWt4hFRU9yOQ_NhK)** ([raw file](https://github.com/hammerlab/pyensembl/blob/4e3f9610b159b0056a153b201b142695e9f5e530/pyensembl/memory_cache.py#L67)):</sup>
This probably warrants a comment re why `chunksize` and why that particular `chunksize`.

---

<sup>**[test/common.py, line 10 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/119#-JxlOMi1LXJLWhDxIW5n)** ([raw file](https://github.com/hammerlab/pyensembl/blob/4e3f9610b159b0056a153b201b142695e9f5e530/test/common.py#L10)):</sup>
Lame!

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pyensembl/119)

tavinathanson:
Interesting choice to add `pylint` to conda rather than `requirements.txt`.  I guess that makes sense; maybe I'll switch the other repos to do the same?

Otherwise, this looks good. A few questions/comments, but nothing major. (+ waiting for Travis to actually pass)

---

Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pyensembl/119)

tavinathanson:
(I might be wrong about needing to re-encrypt the PyPi password, since I added that password a while ago and I may have encrypted it properly)

tavinathanson:
Reviewed 4 of 8 files at r2, 13 of 13 files at r3.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

<sup>**[lint.sh, line 8 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/119#-JxpQZc9ttkBxhbDmpNH)** ([raw file](https://github.com/hammerlab/pyensembl/blob/73387b19af055d1bf045617c276543708dc26939/lint.sh#L8)):</sup>
Can you put this in `pylintrc` so we have one place for all our pylint exceptions?

---

<sup>**[pyensembl/genome.py, line 1030 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/119#-JxpQfyfyANRvH_bALOh)** ([raw file](https://github.com/hammerlab/pyensembl/blob/73387b19af055d1bf045617c276543708dc26939/pyensembl/genome.py#L1030)):</sup>
Test?

---

<sup>**[pyensembl/gtf_parsing.py, line 224 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/119#-JxlRGkUXZXMT-UVMlN7)** ([raw file](https://github.com/hammerlab/pyensembl/blob/4e3f9610b159b0056a153b201b142695e9f5e530/pyensembl/gtf_parsing.py#L224)):</sup>
I think it's a little scary to put off critical tests. Maybe coveralls would help keep us in line. Until then, I hope we get to #120 soon.

---

<sup>**[pyensembl/gtf_parsing.py, line 238 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pyensembl/119#-JxlRv2vsXMsV8Zc3UQS)** ([raw file](https://github.com/hammerlab/pyensembl/blob/4e3f9610b159b0056a153b201b142695e9f5e530/pyensembl/gtf_parsing.py#L238)):</sup>
Is that not the difference between `info` and `debug`?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pyensembl/119)

PR #122, opened 2015-08-28
Deploy to PyPi on every master commit

tavinathanson:
The last deploy failed because it wasn't tagged

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/122)


tavinathanson:
Merging this in because we've talked about it.

PR #134, opened 2016-02-19
Increase the max Ensembl version to 83

iskandr:
...now we wait for tests to fail.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/134)


tavinathanson:
LGTM pending Travis

PR #135, opened 2016-02-22
Fix Ensembl release 83

iskandr:
* make sure all the unit tests work with release 83 and/or use the correct release for which they were written 
* added a few unit tests to improve coverage 
* got rid of the top-level objects like `ensembl77`, `ensembl83`, &c since it seems like bad design to add another one of those for each Ensembl release

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/135)


pyensembl/__init__.py
```diff
@@ -57,12 +58,31 @@ def genome_for_reference_name(reference_name):
...
-ensembl77 = cached_release(77)
-ensembl78 = cached_release(78)
-ensembl79 = cached_release(79)
-ensembl80 = cached_release(80)
-ensembl81 = cached_release(81)
-ensembl82 = cached_release(82)
-ensembl83 = cached_release(83)
-ensembl_grch38 = ensembl83  # most recent for GRCh38
+
+__all__ = [
```
tavinathanson: I don't know much about `__all__` other than some brief Googling; any particular reason you decided to add this now?

test/test_exon_object.py
```diff
@@ -25,8 +27,8 @@ def test_exon_object_by_id():
 
 def test_exon_object_by_id_on_negative_strand():
     """
-    test_exon_object_by_id : check properties of exon 1 from CXCR3 when looked
-    up by ID in Ensembl 77.
+    test_exon_object_by_id_on_negative_strand : check properties of exon 1
```
tavinathanson: Curious why you're just repeating the name of the test?

tavinathanson:
LGTM % minor questions

test/test_exon_object.py
iskandr: It's to combat something annoying that nose does when you're running tests. If a test has a docstring then the name nose shows gets replaced by the docstring, which then obscures which test it's running.

pyensembl/__init__.py
iskandr: I thought it might be slightly better package etiquette to restrict which things get imported as '*'. I'm not totally sure if it's necessary though.

https://docs.python.org/2/tutorial/modules.html#importing-from-a-package

tavinathanson: :+1:

PR #136, opened 2016-02-23
Transcript.coding_sequence didn't work for non-coding transcripts

iskandr:
* Transcript.coding_sequence should return None rather than trying to slice into an empty `Transcript.sequence`

* It was impossible to look up sequences in Ensembl 83 since they added a version number to every FASTA key (added a test for this)

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/136)

### DIFF
--- a/pyensembl/biotypes.py
+++ b/pyensembl/biotypes.py
@@ -282,6 +282,8 @@ long_noncoding = {
     'known_ncrna',
     # unspliced lncRNAs that are several kb in size.
     'macro_lncRNA',
+    # seems to have been added around Ensembl 81
+    'bidirectional_promoter_lncrna',
 }
 
 mitochondrial = {
--- a/pyensembl/sequence_data.py
+++ b/pyensembl/sequence_data.py
@@ -82,7 +82,12 @@ class SequenceData(object):
         fasta_dictionary = {}
         sequence_type = self.sequence_type
         for seq_entry in read(self.fasta_path, format="fasta"):
-            seq_id = seq_entry.metadata["id"]
+            # annoyingly Ensembl83 reformatted the transcript IDs of its
+            # cDNA FASTA to include sequence version numbers
+            # .e.g.
+            # "ENST00000448914.1" instead of "ENST00000448914"
+            # So now we have to parse out the identifier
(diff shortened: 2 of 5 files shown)


pyensembl/sequence_data.py
```diff
@@ -82,7 +82,12 @@ def _parse_fasta_dictionary(self):
         fasta_dictionary = {}
         sequence_type = self.sequence_type
         for seq_entry in read(self.fasta_path, format="fasta"):
-            seq_id = seq_entry.metadata["id"]
+            # annoyingly Ensembl83 reformatted the transcript IDs of its
+            # cDNA FASTA to include sequence version numbers
+            # .e.g.
+            # "ENST00000448914.1" instead of "ENST00000448914"
+            # So now we have to parse out the identifier
+            seq_id = seq_entry.metadata["id"].split(".")[0]
```
tavinathanson: Can you add a test that would break if you didn't have this?

pyensembl/biotypes.py
```diff
@@ -282,6 +282,8 @@
     'known_ncrna',
     # unspliced lncRNAs that are several kb in size.
     'macro_lncRNA',
+    # seems to have been added around Ensembl 81
+    'bidirectional_promoter_lncrna',
```
tavinathanson: Remind me what happens/breaks if you don't have this?

pyensembl/sequence_data.py
iskandr: I did, check out `test_transcript_sequence_ensembl83`

pyensembl/biotypes.py
iskandr: If we search of a Transcript whose biotype isn't in these lists then we get an error. 

Totally open to doing away with this "validation", since Ensembl keeps expanding the lists of valid biotypes.

pyensembl/sequence_data.py
tavinathanson: Oops, somehow didn't think that would cover it, cool

pyensembl/biotypes.py
tavinathanson: :+1: Fine for now

tavinathanson:
LGTM

PR #137, opened 2016-03-23
Add pickling for several classes

tavinathanson:
This will allow `varcode.EffectCollection` to get pickled.

Threw in equals and hashing for species because that was needed for the pickling tests to run.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/137)

### DIFF
--- a/pyensembl/gene.py
+++ b/pyensembl/gene.py
@@ -46,6 +46,7 @@ class Gene(Locus):
 
         Locus.__init__(self, contig, start, end, strand)
 
+        self.require_valid_biotype = require_valid_biotype
         if require_valid_biotype and not is_valid_biotype(biotype):
             raise ValueError(
                 "Invalid gene_biotype %s for gene with ID = %s" % (
@@ -74,14 +75,12 @@ class Gene(Locus):
         return hash(self.id)
 
     def __getstate__(self):
-        fields = self.__dict__.copy()
-        # We can't pickle connections
-        del fields["db"]
-        return fields
+        # Must be in order of __init__ arguments
+        return [self.id, self.name, self.contig, self.start, self.end, self.strand,
+                self.biotype, self.genome, self.require_valid_biotype]
 
     def __setstate__(self, fields):
-        self.__dict__ = fields
-        self.db = self.genome.db
+        self.__init__(*fields)
 
     @memoized_property
     def transcripts(self):
--- a/pyensembl/gene.py
+++ b/pyensembl/gene.py
@@ -73,6 +73,16 @@ class Gene(Locus):
     def __hash__(self):
         return hash(self.id)
 
+    def __getstate__(self):
+        fields = self.__dict__.copy()
+        # We can't pickle connections
+        del fields["db"]
+        return fields
+
+    def __setstate__(self, fields):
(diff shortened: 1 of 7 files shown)


tavinathanson:
@iskandr Don't merge this yet, even if it looks good; I think there's a problem with pickling too much of the `Genome`.

pyensembl/genome.py
```diff
@@ -1041,3 +1042,25 @@ def protein_ids(self, contig=None, strand=None):
...
+            "reference_name",
+            "annotation_name",
+            "annotation_version",
+            "_gtf_path_or_url",
+            "_transcript_fasta_path_or_url",
+            "_protein_fasta_path_or_url",
+            "decompress_on_download",
+            "copy_local_files_to_cache",
+            "require_ensembl_ids",
+            "cache_directory_path"]
```
iskandr: Not sure if this is the right way to go...let me read about the other ways to adding pickling to an object so that we can reconstruct a Genome using fewer hard-coded fields.

tavinathanson: Yeah, fair enough. Maybe I'll do what I did elsewhere and just remove the bad ones vs. add the good ones.

tavinathanson: Ah, right: the annoying thing with `Genome` is that there are a ton of memoized fields that I don't want to save, and `memoized-property` doesn't provide an easy way to get at all of those. What do you think about rolling our own `memoized-property` that lets us easily access all of them and ignore them here?

iskandr: For posterity, we both just learned about `__reduce__`: https://docs.python.org/3/library/pickle.html#object.__reduce__

tavinathanson:
@iskandr Tried to simplify as much as possible; take a look?

pyensembl/gene.py
```diff
@@ -73,6 +74,14 @@ def __eq__(self, other):
     def __hash__(self):
         return hash(self.id)
 
+    def __getstate__(self):
+        # Must be in order of __init__ arguments
+        return [self.id, self.name, self.contig, self.start, self.end, self.strand,
```
iskandr: I don't know if this is easier or more maintainable, but you can also return a dictionary here (which at least lets us later reorder constructor args). Not a big deal though.

pyensembl/gene.py
```diff
@@ -73,6 +74,14 @@ def __eq__(self, other):
...
+                self.biotype, self.genome, self.require_valid_biotype]
+
+    def __setstate__(self, fields):
+        self.__init__(*fields)
```
iskandr: I guess that's an easy solution!

pyensembl/gene.py
tavinathanson: With a dictionary, we start getting into the mapping between `id` and `transcript_id`, which sounded confusing to me.

iskandr: Maybe we should just rename args to match the property? (I know that's a bigger change so maybe for a later PR where we'll also change serialization to use dicts)

tavinathanson: Yeah, that's a fairly annoying change.

iskandr:
LGTM

Issue #150, opened 2016-06-02
Memory Leak / Caching without limits


nosarcasm:
When doing tasks, each action requiring access to the pickled ensembl DB data (e.g looking up a new gene) adds values to the stack causing memory usage to skyrocket. Does not go away after deleting references to pyensembl (e.g deleting the EnsemblRelease variable).

iskandr:
Hi @nosarcasm, 

There's a lot of potentially over-eager caching going on in PyEnsembl:

1) Did you create the `EnsemblRelease` instance yourself or are you using something like  `pyensembl.ensembl_grch38` or [cached_release](https://github.com/hammerlab/pyensembl/blob/master/pyensembl/__init__.py#L42). In the latter case the instance will persist in a global dictionary. 

2) Any call to a method that returns all of some feature (either across the reference or on a contig) such as `gene_names`, `transcript_ids` will cache its results (see: https://github.com/hammerlab/pyensembl/blob/master/pyensembl/genome.py#L365). This might be unnecessary and I'm open to changing it. 

Can you tell me more about what you're doing and how much memory consumption you encounter?

nosarcasm:
I created the ensembl release variable myself. i'm using it in combination with varcode and basically this causes memory usage to go up as each new gene is accessed:

`'''process pseudocode'''
es=EnsemblRelease(75)
for variant in fp:
     variant = varcode.Variant(chrom, pos, ref, alt, ensembl=es)
     effect_on_gene = variant.effects()
     # do something
`
after about 100 genes or so I'm seeing memory usage of 8gb, I usually run out of available memory after about 300 genes (24gb on my virtual machine)

nosarcasm:
I think it may have to do with the memoized properties of transcript.py but this is a complete guess just from looking at the cProfile times and the memory usage spike when accessing variant.effects(), in the sense that memoize is caching objects in memory with each new transcript retrieval but is not forgetting them because the memory is stored in the module variable for memoized_property.py, not in pyensembl. A fix for it would be to offer a monkeypatch for memoized_property.py where you can set a max size for variables in the module or a dealloc function.

iskandr:
Property memoization is definitely a memory leak but I'm surprised by the magnitude. How many variants are you getting effects for? I see >1gb memory usage after about 9k variants. I got rid of some of the memoization in Varcode and now get to >1gb after about 14k variants. 

I'll file another issue under Varcode and try to trim the memory usage of both libraries.

[...]

PR #151, opened 2016-06-07
Getting rid of memoization on many properties and methods

iskandr:
Partially addresses: https://github.com/hammerlab/pyensembl/issues/150

We're still leaking an unbounded amount of memory to memoization but now it's at least being centralized through a few "important" methods such as those which get Gene, Transcript, and Exon objects by ID or the database query helpers.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/151)


pyensembl/gtf.py
```diff
@@ -55,7 +55,7 @@ def __init__(self, gtf_path, cache_directory_path=None):
 
     def __eq__(self, other):
         return (
-            isinstance(other, GTF) and
+            (other.__class__ == GTF) and
```
timodonnell: not important, but just noting I think this change means you can't subclass GTF and have equality work. Not sure if that's intentional

pyensembl/sequence_data.py
```diff
@@ -74,7 +74,7 @@ def __contains__(self, sequence_id):
 
     def __eq__(self, other):
         return (
-            isinstance(other, SequenceData) and
+            (other.__class__ is SequenceData) and
```
timodonnell: ditto to above (I think `is` is better than `==` tho)

timodonnell:
LGTM

pyensembl/gtf.py
iskandr: Hard to imagine when we'd want a sub-class of GTF to equal to a GTF instance, seemed safer to restrict equality to exactly the same class. .

Issue #145, opened 2016-05-10
Specifying install location 


rpadmanabhan:
Hi,
Is there a way to specify a prefix when using pyensembl install to install the ensembl data to a particular directory.

Thank You

iskandr:
@rpadmanabhan 

It's not currently possible to do that in a sane way but is sort-of possible if you don't mind reaching into the internals of pyensembl: you can pass a `cache_directory_path` argument when building a [Genome](https://github.com/hammerlab/pyensembl/blob/master/pyensembl/genome.py#L52). Unfortunately, we never exposed that as an option in [EnsemblRelease](https://github.com/hammerlab/pyensembl/blob/master/pyensembl/ensembl_release.py). 

I'll fix this for you in the next few days (at least from the Python interface for `EnsemblRelease`).

rpadmanabhan:
ah i see , i will try messing around with it and see if it works for me.  Thanks

PR #156, opened 2016-07-21
Override cache dir if PYENSEMBL_CACHE_DIR is set

armish:
so that we can write the cache to a shared directory when running analyses on multi-node clusters.

Context: https://github.com/hammerlab/biokepi/pull/268

cc: @arahuja and @seb

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/156)

### DIFF
--- a/pyensembl/download_cache.py
+++ b/pyensembl/download_cache.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from os import listdir, remove
+from os import listdir, remove, environ
 from os.path import join, exists, split, abspath
 from shutil import copy2, rmtree
 import logging
@@ -20,6 +20,7 @@ import logging
 import datacache
 
 CACHE_BASE_SUBDIR = "pyensembl"
+CACHE_DIR_ENV_KEY = "PYENSEMBL_CACHE_DIR"
 
 def cache_subdirectory(
         reference_name=None,
@@ -111,8 +112,10 @@ class DownloadCache(object):
                 annotation_name=annotation_name,
                 annotation_version=annotation_version)
 
+            # If `CACHE_DIR_ENV_KEY` is set, the cache will be saved there
             self._cache_directory_path = datacache.get_data_dir(
-                subdir=self.cache_subdirectory)
+                subdir=self.cache_subdirectory,
+                envkey=CACHE_DIR_ENV_KEY)
 
         self.decompress_on_download = decompress_on_download
         self.copy_local_files_to_cache = copy_local_files_to_cache
--- a/test/test_download_cache.py
+++ b/test/test_download_cache.py
@@ -1,11 +1,16 @@
 from __future__ import absolute_import
 
-from nose.tools import assert_raises
+from nose.tools import assert_raises, ok_
 from pyensembl.download_cache import (
     DownloadCache,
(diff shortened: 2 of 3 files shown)


pyensembl/download_cache.py
```diff
@@ -111,8 +112,12 @@ def __init__(
                 annotation_name=annotation_name,
                 annotation_version=annotation_version)
 
-            self._cache_directory_path = datacache.get_data_dir(
-                subdir=self.cache_subdirectory)
+            if environ.get(ENV_KEY) is None:
```
arahuja: Isn't the `if` handled by `datacache`?

https://github.com/hammerlab/datacache/blob/master/datacache/common.py#L29-L33

armish: Correct — but I wanted to make this more explicit as it is not apparent from the function call what will happen if you pass both variables.

Actually, maybe adding a comment there to explain this might be better. Let me do that :+1:

armish: done

pyensembl/download_cache.py
```diff
@@ -111,8 +112,10 @@ def __init__(
                 annotation_name=annotation_name,
                 annotation_version=annotation_version)
 
+            # If ENV_KEY is set, the cache will be saved there
             self._cache_directory_path = datacache.get_data_dir(
-                subdir=self.cache_subdirectory)
+                    subdir=self.cache_subdirectory,
```
timodonnell: nit: indentation

test/test_download_cache.py
```diff
@@ -25,3 +30,13 @@ def test_download_cache_missing_remote_file():
...
             path_or_url="ftp://NOTAURL.NOTAURL.NOTAURL")
+
+def test_download_cache_custom_location():
+    test_file = "refseq.ucsc.small.gtf"
+    tmp_dir = tempfile.gettempdir()
+    os.environ['PYENSEMBL_CACHE_DIR'] = tmp_dir
+    download_cache.delete_cache_directory()
+    download_cache.download_or_copy_if_necessary(
+        path_or_url=data_path(test_file))
+    os.path.exists(os.path.join(tmp_dir, test_file))
```
timodonnell: this doesn't test anything, is that intentional?

tavinathanson:
LGTM!

pyensembl/download_cache.py
```diff
@@ -12,14 +12,15 @@
...
-from os import listdir, remove
+from os import listdir, remove, environ
 from os.path import join, exists, split, abspath
 from shutil import copy2, rmtree
 import logging
 
 import datacache
 
 CACHE_BASE_SUBDIR = "pyensembl"
+ENV_KEY = "PYENSEMBL_CACHE_DIR"
```
timodonnell: nit: might call this something a bit clearer and non-abbreviated like `CACHE_DIRECTORY_ENVIRONMENT_VARIABLE`

test/test_download_cache.py
armish: oops - forgot that assert :\

armish:
thanks all! Addressed your comments in my amended commit. Will merge and release on 💚

iskandr:
Fixes https://github.com/hammerlab/pyensembl/issues/145

PR #158, opened 2016-07-25
WIP: Testing builds within Docker images

armish:
Experimental purposes only. Do not merge yet.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/158)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ language: python
 services:
     - docker
 script:
+    - docker -help
+    - dcker -v
     - env
     - echo $TRAVIS_JOB_ID
     - docker build --build-arg TRAVIS_JOB_ID=$TRAVIS_JOB_ID --build-arg PYTHON_VERSION=3 -t "ensembl-test3" .
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,15 @@
 sudo: required
 language: python
 services:
-    - docker
+  - docker
+before_install:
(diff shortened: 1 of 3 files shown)


iskandr:
@armish I've seen some other folks name unfinished branches like "WIP: Testing builds within Docker images" and then remove the "WIP" once you're done. Not sure if that alleviates worries about a premature merge but I've started doing it.

iskandr:
Did this end up not working out?

armish:
@iskandr: yeah, it came with some other complications that I haven't anticipated. @ryan-williams's recent suggestion about using Travis's cache mechanism sounds promising, though.

tavinathanson:
What was @ryan-williams's suggestion about using Travis's cache mechanism? We had earlier talked about using that to cache the raw Ensembl data, however it would still need to be downloaded locally to the Travis node running the tests, and then the DB would still need to be built.

Maybe @ryan-williams's suggestion was to cache the DB itself via Travis's cache? We initially were hesitant to do that since the DB might keep changing, but I guess we did warm up to that idea in conversations about this Docker image approach?

iskandr:
@ryan-williams @armish @tavinathanson If we add the database schema version to the path and then cache on Travis, then maybe (fingers crossed) it would be OK?

tavinathanson:
@iskandr interesting idea! Other than forgetting to update the schema version, I don't see any obvious issues there?

PR #163, opened 2016-09-14
Deserialization should construct cached EnsemblRelease objects

iskandr:
* Moved contents of `cached_release` onto the EnsemblRelease object (`EnsemblRelease.cached`)
* Using `WeakValueDictionary` instead of ordinary dict to avoid memory leak
* Added `from_dict` class method to use `cached` for deserialization

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/163)

### DIFF
--- a/pyensembl/__init__.py
+++ b/pyensembl/__init__.py
@@ -43,7 +43,7 @@ def cached_release(release, species="human"):
     otherwise returns the old instance.
 
     Keeping this function for backwards compatibility but this functionality
-    has been moving onto the EnsemblRelease object itself by overriding __new__.
+    has been moving into the cached method of EnsemblRelease.
     """
     return EnsemblRelease.cached(release=release, species=species)
 
--- a/pyensembl/__init__.py
+++ b/pyensembl/__init__.py
@@ -35,7 +35,7 @@ from .species import (
 )
 from .transcript import Transcript
 
-__version__ = '0.9.6'
+__version__ = '0.9.7'
 
 def cached_release(release, species="human"):
     """
--- a/pyensembl/__init__.py
+++ b/pyensembl/__init__.py
@@ -17,7 +17,7 @@ from __future__ import print_function, division, absolute_import
 from .memory_cache import MemoryCache
 from .download_cache import DownloadCache
 from .ensembl_release import EnsemblRelease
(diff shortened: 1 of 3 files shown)


iskandr:
It would be nice to do something similar for the `Genome` base class but I would wait until we have a more automatic way of adding this functionality to a class (maybe via a metaclass or modification to the `Serializable` base class?)

pyensembl/__init__.py
```diff
@@ -35,28 +35,23 @@
...
+    """
+    Create an EnsemblRelease instance only if it's hasn't already been made,
+    otherwise returns the old instance.
 
-def cached_release(version, species="human"):
-    """Cached construction of EnsemblRelease objects. It's desirable to reuse
-    the same EnsemblRelease object since each one will store a lot of cached
-    annotation data in-memory.
+    Keeping this function for backwards compatibility but this functionality
+    has been moving onto the EnsemblRelease object itself by overriding __new__.
```
tavinathanson: Where are you overriding new?

pyensembl/ensembl_release.py
```diff
@@ -32,18 +33,53 @@ class EnsemblRelease(Genome):
...
+            server=ENSEMBL_FTP_SERVER):
+        """
+        Construct EnsemblRelease if it's never been made before, otherwise
+        return an old instance.
+        """
+        init_args_tuple = cls.normalize_init_values(release, species, server)
+        if init_args_tuple in cls._genome_cache:
+            genome = cls._genome_cache[init_args_tuple]
+        else:
+            genome = cls._genome_cache[init_args_tuple] = cls(*init_args_tuple)
```
tavinathanson: Didn't even know this was valid syntax! It's right to left?

tavinathanson:
Agree that we need to generalize this, but good to have a solution for now!

pyensembl/__init__.py
iskandr: Sorry, that's left over from an old attempt (turned out too arcane).

pyensembl/ensembl_release.py
iskandr: Yep, right to left (like Arabic!). It's a generalization of `a = b = None`.

tavinathanson:
LGTM

PR #174, opened 2016-10-11
Fix logging file access in shell (needed pkg_resources)

julia326:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/174)


ihodes:
LGTM

PR #185, opened 2017-05-04
Add non-coding RNA sequences to pyensembl

murphycj:
Pyensembl now downloads the ncRNA FASTA file so the transcript sequence is available

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pyensembl/185)


murphycj:
Working on getting the unit tests to play nice. But I guess I can see one issue with the approach I took. It seems pyensembl expects just one FASTA file for transcripts. My patch easily enough accommodates the ncRNA FASTA from ensembl, but I think my solution would not play well for people who are specifying custom GTFs and FASTA files. Maybe then a more general solution is to have pyensembl accept an arbitrary amount of transcript FASTA files?

iskandr:
I think I prefer the solution of allowing a variable number of FASTA files for RNA sequences. @tavinathanson curious to hear you opinion.

iskandr:
I think I also need to understand why Ensembl splits up the `cdna.all` FASTA from the `ncrna`.

pyensembl/genome.py
```diff
@@ -311,6 +350,9 @@ def install_string(self):
         if self.has_transcript_fasta:
             args.append("--transcript-fasta")
             args.append("\"%s\"" % self._transcript_fasta_path_or_url)
+        if self.has_ncrna_fasta:
+            args.append("--nrcna-fasta")
```
tavinathanson: Typo here: `nrcna`

test/test_mouse.py
```diff
@@ -18,6 +18,10 @@ def test_mouse_ENSMUSG00000017167():
     fasta/mus_musculus/cdna/Mus_musculus.GRCm38.cdna.all.fa.gz via:
     grep "ENSMUSG00000017167" Mus_musculus.GRCm38.cdna.all.fa -A 50
 
+    ncRNA FASTA cropped from ftp://ftp.ensembl.org/pub/release-81/
+    fasta/mus_musculus/cdna/Mus_musculus.GRCm38.ncrna.fa.gz via:
```
tavinathanson: This URL doesn't look correct? The `ncrna` file isn't under the `cdna` folder in reality, right?

murphycj: Indeed it is not correct. The correct one is: ftp://ftp.ensembl.org/pub/release-81/fasta//mus_musculus/ncrna/Mus_musculus.GRCm38.ncrna.fa.gz

tavinathanson:
@iskandr @murphycj I'm inclined to agree that a variable number of RNA `FASTA` files seems cleaner both for the code and for users.

murphycj:
I added functionality so that pyensembl accepts an arbitrary number of transcript fasta files. It also still downloads the ncRNA file. Here is an example usage of the command line:

    pyensembl install --reference-name "test" --annotation-name test --gtf test.gtf --transcript-fasta test1.fa --transcript-fasta test2.fa

pyensembl/genome.py
```diff
@@ -66,8 +66,8 @@ def __init__(
         gtf_path_or_url : str
             Path or URL of GTF file
 
-        transcript_fasta_path_or_url : str
-            Path or URL of FASTA file containing transcript sequences
+        transcript_fasta_paths_or_urls : str
```
iskandr: Is this still a string or a list of strings?

pyensembl/genome.py
```diff
@@ -169,17 +169,19 @@ def _get_gtf_path(self, download_if_missing=False, overwrite=False):
...
-            path_or_url=self._transcript_fasta_path_or_url,
-            download_if_missing=download_if_missing,
-            overwrite=overwrite)
+        return [
+            self._get_cached_path(
+                field_name="transcript-fasta",
+                path_or_url=path,
+                download_if_missing=download_if_missing,
+                overwrite=overwrite
+            ) for path in self._transcript_fasta_paths_or_urls]
```
iskandr: Really small syntax nit but I think it's more idiomatic to move the closing paren to the previous line.

pyensembl/genome.py
```diff
@@ -438,7 +442,11 @@ def transcript_sequence(self, transcript_id):
         if self.transcript_sequences is None:
             raise ValueError(
                 "No transcript FASTA supplied to this Genome: %s" % self)
-        return self.transcript_sequences.get(transcript_id)
+        sequence = [x for x in map(lambda x: x.get(transcript_id), self.transcript_sequences) if x is not None]
+        #self.transcript_sequences.get(transcript_id)
```
iskandr: Stray commented out code?

pyensembl/genome.py
```diff
@@ -438,7 +442,11 @@ def transcript_sequence(self, transcript_id):
         if self.transcript_sequences is None:
             raise ValueError(
                 "No transcript FASTA supplied to this Genome: %s" % self)
-        return self.transcript_sequences.get(transcript_id)
+        sequence = [x for x in map(lambda x: x.get(transcript_id), self.transcript_sequences) if x is not None]
```
iskandr: It might make more sense to combine the transcript sources when we're constructing `transcript_sequences` and check for duplicates then. 

Also, style-wise, the double usage of `x` made this particularly tricky to parse.

pyensembl/transcript.py
```diff
@@ -400,7 +400,10 @@ def sequence(self):
         Spliced cDNA sequence of transcript
         (includes 5" UTR, coding sequence, and 3" UTR)
         """
-        return self.genome.transcript_sequences.get(self.id)
+        for transcript_sequences in self.genome.transcript_sequences:
+            if self.id in transcript_sequences:
+                return transcript_sequences.get(self.id)
```
iskandr: If we do keep multiple dictionaries then at this point you can get away with `transcript_sequences.[self.id]` but I'm still leaning toward combining all the dictionaries up front.

murphycj: Hm, I agree that combining all transcripts into one dictionary may be a better solution. How then to store the pickled versions of the transcript FASTA files? Combine them into something like transcripts.pickle?

murphycj: Ah, I guess one issue with combining everything into transcripts.pickle is backwards compatibility. The change would require everyone to "re-pickle" their transcript and protein sequences.

murphycj:
So I made the change so all transcript sequences are combined into one dictionary. However, there is one odd issue I noticed when running the following code:

    import pyensembl
    db = pyensembl.EnsemblRelease(84,'mouse')
    g = db.gene_by_id('ENSMUSG00000022770')
    t1 =g.transcripts[0]
    t1.sequence

I get the following message after that last line (in addition to it printing the sequence)

    No handlers could be found for logger "pyensembl.sequence_data"

pyensembl/genome.py
murphycj: it is a list of strings

pyensembl/genome.py
murphycj: This was intentional. My latest code went back to how this code section originally was, though

iskandr:
@julia326 Do you have any idea what's going on with logging here? Have you seen that warning before?

murphycj:
@iskandr  After my latest commit, I don't see the same logging issue anymore... not sure what happened there. Maybe just something weird how my machine was configured (e.g. did not run pip uninstall pyensembl to clean up things)

iskandr:
@murphycj Sorry this PR fell off my radar, do you think it's done?

murphycj:
@iskandr No problem. Not quite done. Been busy, but have made a few more changes (features for multiple protein FASTA files and GTF files). I'll finish them up within the next couple days and commit them.

pyensembl/sequence_data.py
```diff
@@ -36,22 +37,24 @@ class SequenceData(object):
     """
     def __init__(
             self,
-            fasta_path,
+            fasta_paths,
             require_ensembl_ids=False,
             cache_directory_path=None):
```
iskandr: In the interest of not breaking any other existing code which may use `SequenceData(fasta_path)`, do you mind adding:
```
if isinstance(fasta_paths, six.string_types):
    fasta_paths = [fasta_paths]
```

murphycj:
@iskandr So, I added feature to handle multiple protein fasta files as well. Might be enough for now. Handling multiple GTF files seems a bit more complicated. Seems there are two main solutions. (1) have pyensembl merge all GTF files into one (so creating static names like gtf.db, gtf.expanded.csv, etc. in the cache folder)--but this option would break backwards compatibility. (2) Keep separate .db, .expanded.csv, etc. files for each GTF file, then just keep multiple sqlite database connections to each of them--this would keep backwards compatibility.

iskandr:
I think limiting this branch to FASTA files seems fine to me, especially when the union of the identifiers in the FASTA files is a subset of those found in the single GTF file.

murphycj:
Ok, we can leave the multiple GTF patch for a later date. Merge this branch then if it suits you.

pyensembl/sequence_data.py
```diff
@@ -36,22 +38,29 @@ class SequenceData(object):
...
+
+        if isinstance(fasta_paths, string_types):
+            fasta_paths = [fasta_paths]
+
+        self.fasta_paths = [abspath(path) for path in fasta_paths]
+        self.fasta_directory_paths = [split(path)[0] for path in self.fasta_paths]
+        self.fasta_filenames = [split(path)[1] for path in self.fasta_paths]
         if cache_directory_path:
-            self.cache_directory_path = cache_directory_path
+            self.cache_directory_paths = [cache_directory_path]*len(self.fasta_paths)
```
murphycj: I will fix this. I do think it would be a cleaner solution to just put all files into one cache directory. 

It seems the user cannot specify the cache directory (unless they are interfacing with the API directly in a script), so it would need to be automatically determined. But the way things are done currently, the cache directory (if it not supplied through `cache_directory_path `) is determined by the paths of the supplied FASTA files. And the case may arise where the user provides multiple transcript or protein FASTA files where each is located in separate directories, so that I why I used a list for `cache_directory_paths`.

iskandr: I see -- there's something slightly unsettling about having to keep zippable lists of prefix dirs and filenames but I also can't think of a better solution.

murphycj: I think the only two alternative solutions are to automatically create the cache directory in global cache directory supplied from datacache python package (e.g. `/Users/charlesmurphy/Library/Caches/pyensembl` on my Mac) or force the user to always specify `cache_directory_paths` when working with their own custom FASTA files. If neither solution sounds good, then keep the current solution.

PR #196, opened 2018-01-17
Fixing travis

iskandr:
Problem: changes to Travis seem to have limited us to <4gb / 19gb of disk space. Our unit tests use significantly more than that while installing multiple Ensembl releases. 

Solution:
* Tell Travis to install fewer things using `language: minimal` in `.travis.yml`. 
* Don't install `scipy` or `matplotlib`
* Don't install Ensembl releases 54 or 83
* Don't index every table on `['seqname', 'start', 'end', 'strand']`, since we're already doing `['seqname', 'start', 'end']` and if there are query results on both strands they can probably be efficiently filtered.
* Got rid of `GTF` object, which existed to cache query results on disk as CSV files. This slightly improved long-term performance at the cost of massive disk usage. 

During my hunt for disk savings I also added a `delete-source-files` command to the `pyensembl` CLI which allows you delete `GTF` and `FASTA` files while leaving their indexed representations in place.


pyensembl/database.py
```diff
@@ -39,47 +41,57 @@ class Database(object):
...
-    def local_db_path(self):
-        if not self.gtf:
-            raise ValueError("No GTF supplied to this Database: %s" %
-                             str(self))
-        dirpath = self.gtf.cache_directory_path
-        filename = self.local_db_filename()
-        return join(dirpath, filename)
+        return hash((self.gtf_path))
+
+    @property
```
julia326: rename seems unnecessary for this PR, change it in a later one to reduce delta

pyensembl/genome.py
```diff
@@ -374,6 +371,19 @@ def clear_cache(self):
             if hasattr(maybe_fn, "clear_cache"):
                 maybe_fn.clear_cache()
 
+    def delete_source_files(self):
```
julia326: i'd get rid of it in this PR, not used and a bit hackish

pyensembl/shell.py
```diff
@@ -55,72 +55,81 @@
...
-        "--protein-fasta",
-        type=str,
-        default=None,
-        help="URL or local path to a FASTA file containing protein data.")
-
-    parser.add_argument("action",
-        type=lambda arg: arg.lower().strip(),
-        choices=("install", "delete-all-files", "delete-index-files"),
-        help="\"install\" will download and index any data that is  not "
+parser = argparse.ArgumentParser(usage=__doc__)
```
julia326: this file can be reverted, since it's only related to the delete-source-files addition

PR #205, opened 2018-07-27
List dirs in cli

iskandr:


pyensembl/shell.py
```diff
@@ -177,7 +178,11 @@ def run():
     if args.action == "list":
         genomes = collect_all_install_ensembl_releases()
         for genome in genomes:
-            print("-- %s" % genome)
+            # print every directory in which downloaded files are located
+            # in most case this will be only one directory
+            filepaths = genome.required_local_files()
+            directories = {os.path.split(path)[0] for path in filepaths}
+            print("-- %s: %s" % (genome, ", ".join(directories)))
```
julia326: Use logging instead of print statements here?

Issue #209, opened 2018-07-30
Add an exception for Saccharomyces cerevisiae in `is_valid_ensembl_id(ensembl_id)` function


rraadd88:
Hello developers,
I experienced errors when I tried to work with *Saccharomyces cerevisiae* genome using `pyensembl`. I could install the genome successfully. But a function called  `is_valid_ensembl_id(ensembl_id)`  disallows me from doing any more than installing the genome. It is located at
https://github.com/openvax/pyensembl/blob/4e2af0a00b10702fbc6e0ba66bafca79a2031d2c/pyensembl/common.py#L87

I would imagine it would be a quick fix to add an exception for Saccharomyces cerevisiae genome. In case of yeast, the gene ids (unconventionally) start with 'Y' eg. YHR055C, YPR161C, YOL138C etc.

I hope you would be able to incorporate this in the next release.

(Also, I saw that the [current build is failing](https://travis-ci.org/openvax/pyensembl/builds/410073849) which is disconcerting. :/ )
Thank you.

PR #210, opened 2018-07-31
added exception is_valid_id for yeast

rraadd88:
Related to issue [#209](https://github.com/openvax/pyensembl/issues/209): Add an exception for Saccharomyces cerevisiae in `is_valid_ensembl_id(ensembl_id)` function

As you would be able to notice, I have added an `or` argument to allow an exception to yeast ids (`or ensembl_id.startswith("Y")`)
I would like to acknowledge that this quick fix may cause problems if users provide gene names starting with Y (they would not get an error).

Ideally, the species name (taxonomic one) should be provided to the function. Here, I went for a quick fix because otherwise I would have to crawl back to all the places where  `is_valid_{}_id` function is used and fix them systematically.

I hope if the quick fix is ok.   
If not let me know if I should make an effort to fix this issue systematically.

### DIFF
--- a/pyensembl/common.py
+++ b/pyensembl/common.py
@@ -84,7 +84,7 @@ def memoize(fn):
 
 def is_valid_ensembl_id(ensembl_id):
     """Is the argument a valid ID for any Ensembl feature?"""
-    return is_string(ensembl_id) and ensembl_id.startswith("ENS")
+    return is_string(ensembl_id) and (ensembl_id.startswith("ENS") or ensembl_id.startswith("Y"))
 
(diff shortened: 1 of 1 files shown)


iskandr:
Hi @rraadd88 -- this is one of a few different examples that are making me realize that Ensembl gene naming is much less consistent than I previously realized. I think that I'll actually just remove the ID validation, since PyEnsembl already has to codify a ton of fluidly evolving standards about Ensembl's data.

iskandr:
Closing since I got rid of Ensembl ID validation: https://github.com/openvax/pyensembl/pull/211

PR #212, opened 2018-09-17
Add transcript_support_level (TSL) field

vladsaveliev:
GRCh38 database provides a field called `transcript_support_level` (TSL) for transcript features. This PR adds reading of it as an optional field, and adds a new `support_level` field for `Transcript` object.

### DIFF
--- a/test/test_transcript_support_level.py
+++ b/test/test_transcript_support_level.py
@@ -9,14 +9,26 @@ from nose.tools import eq_
 from pyensembl import cached_release
 
 def test_transcript_support_level():
+    """ The Transcript Support Level (TSL) is a method to highlight the well-supported and poorly-supported transcript
+        models for users, based on the type and quality of the alignments used to annotate the transcript.
+        In the Ensembl database, it can be assigned to a value 1 through 5, or reported as NA, or missing, or missing
+        completely in older releases. We translate it to an integer value, otherwise to None.
+    """
     ensembl93 = cached_release(93)
     transcript = ensembl93.transcripts_by_name("DDX11L1-202")[0]
-    eq_(transcript.support_level, "1")
-    # for this transcript, the tsl value is missing, and should yield an empty string:
-    transcript = ensembl93.transcripts_by_name("OR4G11P-202")[0]
-    eq_(transcript.support_level, '')
+    eq_(transcript.support_level, 1)
 
-    # transcript_support_level columns was missing in GRCh37 and older releases of GRCh38, and should yield None:
+    # For this transcript, the transcript_support_level value is missing in the database record:
+    transcript = ensembl93.transcripts_by_name("OR4G11P-202")[0]
+    eq_(transcript.support_level, None)
+
+    # Some features are reported as "NA" in Ensembl: those are features like pseudogenes, single exon transcripts,
+    # HLA, T-cell receptor and Ig transcripts that are not analysed in terms of TSL and therefore not given any
+    # of the TSL categories. We translate NA to None as well.
+    transcript = ensembl93.transcripts_by_name("MIR1302-2-201")[0]
+    eq_(transcript.support_level, None)
+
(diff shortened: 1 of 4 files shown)


iskandr:
Thanks @vladsaveliev -- this looks like a good change. 

I noticed that the Travis build when it tries to use our private netmhcbundle package. This might be a permissions problem, I'll look into it. 

In the mean time, do you mind adding a unit test for this property?

vladsaveliev:
Thanks @iskandr!

Just added a test for this.

iskandr:
Failing on Travis due to this issue: https://github.com/travis-ci/travis-ci/issues/1946 (encrypted values not available on PRs from forks)

iskandr:
Only solution I've seen so far seems pretty complicated: https://blog.algolia.com/travis-encrypted-variables-external-contributions/

iskandr:
OK @vladsaveliev -- if you rebase then Travis should work for your PR.

test/test_transcript_support_level.py
```diff
@@ -0,0 +1,22 @@
...
+from __future__ import absolute_import
+
+from nose.tools import eq_
+
+from pyensembl import cached_release
+
+def test_transcript_support_level():
+    ensembl93 = cached_release(93)
+    transcript = ensembl93.transcripts_by_name("DDX11L1-202")[0]
+    eq_(transcript.support_level, "1")
```
iskandr: Are the support levels always strings? Do they ever contain anything other than "1" ... "5"?

test/test_transcript_support_level.py
```diff
@@ -0,0 +1,22 @@
...
+    # for this transcript, the tsl value is missing, and should yield an empty string:
+    transcript = ensembl93.transcripts_by_name("OR4G11P-202")[0]
```
iskandr: Do you have any idea why/when the support level is missing for transcripts? Also is it desirable to empty strings as a distinct missing value from `None`?

test/test_transcript_support_level.py
vladsaveliev: "1" through "5", or "NA". Because of the latter, hesitated to convert to int.

test/test_transcript_support_level.py
vladsaveliev: I think it might be missing for pseudogenes (and in this case, processed transcripts):

> Transcript Support Level Not Analysed. Pseudogenes, single exon transcripts, HLA, T-cell receptor and Ig transcripts are not analysed and therefore not given any of the TSL categories.

An empty string is what the db returns on a missing value of an existing column. My reasoning was that an empty string allows to distinguish a missing value from a missing column. On the other hand, a use can always check with `db.column_exists()`, so I think I'll just change it to None.

vladsaveliev:
Rebased!

iskandr:
Whew, almost there. Found two failing serialization tests: 

* test.test_serialization.test_json_ensembl_transcript
* test.test_serialization.test_pickle_ensembl_transcript

I'm guessing they're missing the TSL?

vladsaveliev:
Right, because I wasn't adding `support_level` field into the `state_dict` returned by `to_dict()`, that is used by Serializable to serialize to json. Fixed now.

iskandr:
Ok, looks great. Can you bump the version number in __init.py__? Then I'll merge the PR and a successful Travis run on master will deploy to PyPI.

vladsaveliev:
Done!

iskandr:
Merged! Thanks for the PR!

Also, feel free to leave questions on the vaxrank and docker pipeline repos if you encounter anything confusing there.

vladsaveliev:
By the way, thank you for all the tools you guys do in openvax. It's remarkably high quality.

iskandr:
No problem -- our big weakness has so far been documenting and publicizing what we're building. Feel free to ask about anything that's unclear or nudge us to document things better.

PR #220, opened 2019-03-31
Increase max Ensembl version to the current 95

krassowski:


iskandr:
Thanks @krassowski. 

I'll do a later PR to upgrade the unit tests to run on release 95.

krassowski:
@iskandr thank you for maintaining pyensembl!

PR #222, opened 2019-07-22
Updated ensembl_release_versions.py

Flosz:
Current version does not allow to get versions > 95. This allows access to versions 96 & 97.


iskandr:
I'm trying to figure out why external contributors can't run tests on Travis but for now I merged an equivalent PR

PR #226, opened 2019-09-05
Fix chrM / MT normalization

iskandr:
In the dark ages of PyEnsembl we needed a quick way to annotate variants from hg19 and GRCh37 using Ensembl reference data. This lead to a dirty chromosome normalization hack where we turn e.g. "chr1" -> "1" and "chrM" -> "MT". 

This was always a little questionable (since the mitochondrial sequences aren't actually the same) but even worse is incorrect for references like GRCh38. 

So, this PR gets rid of two aspects of contig normaliztion: "chr" prefix is now preserved and we don't convert "M" -> "MT" for the mitochondrial genome. 

Fixes: https://github.com/openvax/pyensembl/issues/225

To restore use of Ensembl data for hg19 variants we'll have to make a more local contig conversion option in Varcode.

### DIFF
--- a/pyensembl/normalization.py
+++ b/pyensembl/normalization.py
@@ -38,17 +38,13 @@ def normalize_chromosome(c):
     elif result == "":
         raise ValueError("Chromosome name cannot be empty")
 
-    # only strip off lowercase chr since some of the non-chromosomal
-    # contigs start with "CHR"
-    if result.startswith("chr"):
-        result = result[3:]
-
-    # just in case someone is being lazy, capitalize "M", "MT", X", "Y"
-    result = result.upper()
-
-    # standardize mitochondrial genome to be "MT"
-    if result == "M":
-        result = "MT"
+    if result.startswith("chr") and "_" not in result:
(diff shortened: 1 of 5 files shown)


pyensembl/normalization.py
```diff
@@ -38,17 +38,13 @@ def normalize_chromosome(c):
...
-    if result.startswith("chr"):
-        result = result[3:]
-
-    # just in case someone is being lazy, capitalize "M", "MT", X", "Y"
-    result = result.upper()
-
-    # standardize mitochondrial genome to be "MT"
-    if result == "M":
-        result = "MT"
+    if result.startswith("chr") and "_" not in result:
```
julia326: thought for the future: normalize_chromosome could be aware of its Pyensembl release and normalize it to that set of contig names?

iskandr: I was thinking something similar! It's a little tricky since the normalization also gets applied to GTF file contigs which are kind of a lower level concept but I suspect we could figure out a sane rewiring.

PR #230, opened 2019-10-18
Add check .start_codon_complete

### LINKED ISSUE #231: Error when checking that transcript is complete in ensembl69
Using release 69 of Ensembl, I ran into an issue where some transcripts existed which had an annotated start_codon that only spanned two positions (example ENST00000543092).

Because of this, when .complete() was run, self.coding_sequence would hit an error when it tried to determine the _codon_positions:
ValueError: Expected 3 positions for start_codon of ENST00000543092 but got 2

It seems that more recent releases of Ensembl do not have this issue (they no longer list a start_codon for these cases).

Using Python 3.7.3 and Isovar 1.0.10.

scottdbrown:
Using release 69 of Ensembl, I ran into an issue where some transcripts existed which had an annotated start_codon that only spanned two positions (example ENST00000543092).

Because of this, when .complete() was run, self.coding_sequence would hit an error when it tried to determine the _codon_positions:
ValueError: Expected 3 positions for start_codon of ENST00000543092 but got 2

It seems that more recent releases of Ensembl do not have this issue (they no longer list a start_codon for these cases).

To continue using Ensembl 69, I added the additional property .start_codon_complete which catches a _codon_position ValueError and returns False, and added self.start_codon_complete as an additional check in .complete().

Fixes #231

### DIFF
--- a/pyensembl/transcript.py
+++ b/pyensembl/transcript.py
@@ -234,6 +234,17 @@ class Transcript(LocusWithGenome):
         stop_codons = self._transcript_feature_position_ranges(
             "stop_codon", required=False)
         return len(stop_codons) > 0
+    
+    @memoized_property
+    def start_codon_complete(self):
+        """
+        Does the start codon span 3 genomic positions?
+        """
+        try:
+            pos = self._codon_positions("start_codon")
+        except ValueError:
+            return False
+        return True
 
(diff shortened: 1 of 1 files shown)


iskandr:
This looks great, OK to merge. Thanks!

PR #249, opened 2021-04-08
Update max ensemble version from 102 -> 103

Jiafi:


ialbert:
Could someone please bump this to 104 and merge it so that the latest releases could be obtained. It would be a shame to lose the utility of all this code, just because the releases are hardcoded.

timodonnell:
Thanks @Jiafi and @ialbert - done

PR #261, opened 2022-01-06
bump ensembl release version

jkasak:
105 is the latest release, while 104 is the currently enforced upper bound in the library.


iskandr:
Thanks!

PR #268, opened 2022-08-18
only split name of ENSEMBL naming

y9c:
In other databases, such as TAIR, the '.1' notation is the isoform not the version.


iskandr:
This is a helpful fix, thanks!

PR #271, opened 2022-11-25
Update ensembl link to download file

pamonlan:
Error downloading ensembl files, used to work with ftp, now with https. Change in ensembl_url_templates.py


timodonnell:
thanks @pamonlan !

PR #277, opened 2022-12-22
Add pandas>=1.3.5 as requirement

deto:
Pandas needed when running pyensembl install and will break if lower than this version.


iskandr:
Thanks @deto! 

I'm working on actually getting rid of Pandas (as well as sqlite3) and migrating everything to use Polars and simpler lookup/index structures. Until then, this is a useful fix!

PR #281, opened 2023-02-15
Update species.py

rraadd88:
Link to the documentation.

### DIFF
--- a/pyensembl/species.py
+++ b/pyensembl/species.py
@@ -145,7 +145,7 @@ def normalize_species_name(name):
 def find_species_by_name(species_name):
     latin_name = normalize_species_name(species_name)
     if latin_name not in Species._latin_names_to_species:
-        raise ValueError("Species not found: %s" % species_name)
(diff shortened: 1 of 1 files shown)


pyensembl/species.py
```diff
@@ -145,7 +145,7 @@ def normalize_species_name(name):
 def find_species_by_name(species_name):
     latin_name = normalize_species_name(species_name)
     if latin_name not in Species._latin_names_to_species:
-        raise ValueError("Species not found: %s" % species_name)
+        raise ValueError("Species not found: %s; For non-Ensembl Data, see https://github.com/openvax/pyensembl#non-ensembl-data" % species_name)
```
iskandr, line 148: Small nit, can you change this to:

```
raise ValueError("Species not found: %s; for non-Ensembl data see https://github.com/openvax/pyensembl#non-ensembl-data" % (species_name,))
```

Also, if you increase the library version to 2.2.7 then it will auto-deploy to PyPI on merge

iskandr:
Added in https://github.com/openvax/pyensembl/pull/282 -- thanks!

PR #302, opened 2024-01-16
Update README.md

rraadd88:
Corrected the missing import of `find_species_by_name`.


iskandr:
I guess that you don't even need to call that function directly, right? Doesn't it get used implicitly if you pass in the species name?

PR #305, opened 2024-03-28
Add support for downloading plant species with pyEnsembl

pamonlan:
This commit introduces the ability to download plant species data using pyEnsembl. We've added the 'is_plant' parameter to the Species class, and registered two new species: Arabidopsis Thaliana and Oryza Sativa (Rice).

We've also added the ENSEMBL_PLANTS_FTP_SERVER URL, and the PLANTS_GTF_SUBDIR_TEMPLATE and PLANTS_FASTA_SUBDIR_TEMPLATE for creating the download links. The code checks if the species is a plant to determine which templates to use.


iskandr (approved):
LGTM!

PR #306, opened 2024-04-08
Update species.py

pamonlan:
Commit message: Change pig synonym to pork to avoid error

In the Species.register function call, replace the 'pig' synonym with 'pork' to prevent the ValueError raised due to synonym conflict.


25shmeckles (commented):
There is a typo in the code. It should be `"pork"` not `"pok"`

Issue #335, opened 2026-05-05
ID version handling and GENCODE compatibility


ccwang002:
Hi team, thanks again for maintaining this library.

We (together with @fkgruber) are trying to use pyensembl with a custom genome build based on GENCODE v48, but we’re running into issues where variant effects are not inferred correctly. For example,

```python
from pyensembl import Genome
from varcode import Variant

g = Genome(
    reference_name="GRCh38",
    annotation_name="my_genome_features",
    gtf_path_or_url="gencode.v48.primary_assembly.annotation.gtf",
    protein_fasta_paths_or_urls="gencode.v48.pc_translations.fa.gz",
    transcript_fasta_paths_or_urls="gencode.v48.transcripts.fa.gz",
    cache_directory_path=".cache",
)

Variant("chr1", 12110124, "G", "A", ensembl=g).effects()
# <EffectCollection with 3 elements>
#  -- NoncodingTranscript(...)
#  -- NoncodingTranscript(...)
#  -- NoncodingTranscript(...)

# Expected output say using Ensembl release 114
# <EffectCollection with 3 elements>
#  -- Substitution(variant=..., transcript_name=TNFRSF8-201, transcript_id=ENST00000263932, effect_description=p.R199H)
#  -- Substitution(variant=..., transcript_name=TNFRSF8-203, transcript_id=ENST00000417814, effect_description=p.R88H)
#  -- NoncodingTranscript(variant=..., transcript_name=TNFRSF8-205, transcript_id=ENST00000514649)
```

So far we've identified some potential issues:

1. Protein/transcript ID version handling. 
    It looks like pyensembl trims version suffixes from FASTA IDs (e.g. `.1` in `ENSPxxx.1`), which seems to break mapping between GTF and FASTA entries.
     - GENCODE GTF includes versioned IDs directly (e.g. `ENSTxxx.7`, `ENSPxxx.1`)
     - Ensembl GTF separates version into a different field (`transcript_version`, `protein_version`)

    The version trimming seems to occur at these places:
     - https://github.com/openvax/pyensembl/blob/v2.6.9/pyensembl/fasta.py#L50-L61
     - https://github.com/openvax/pyensembl/blob/v2.6.9/pyensembl/transcript.py#L454-L463

2. Biotype field naming differences.
    GENCODE uses: `gene_type` `transcript_type`, while pyensembl expects: `gene_biotype`, `transcript_biotype`

[...]

PR #355, opened 2026-05-13
Preserve FASTA-header versions in SequenceData (closes #351)

### LINKED ISSUE #351: Preserve FASTA-header versions in SequenceData instead of stripping at parse time
## Background

Today `pyensembl/fasta.py:50-61` strips ENS `.N` version suffixes from FASTA headers during parsing and discards the version. The dictionary in `SequenceData` therefore keys on the unversioned ID. This was originally added so that GTF `protein_id` / `transcript_id` lookups (which were unversioned in older Ensembl GTFs) would match FASTA headers (which carry the version on more recent Ensembl releases). v2.9.6 paved over the symmetric mismatch on the GENCODE side (where the *GTF* embeds the version) with `lookup_sequence_with_version_fallback` but that helper still has to strip on every call because the version was discarded at parse time.

## Why preserving the version matters

1. **Sequence-identity validation.** If a user `pip install`s an older FASTA (or hand-curates a mix of releases) and `pyensembl.protein_sequence("ENSP00000123456")` returns a sequence, there's no way to confirm the returned bytes correspond to `ENSP00000123456.3` vs `.2` vs `.1`. The version in the FASTA header is the authoritative source-of-truth for the sequence's identity — more authoritative than the GTF's `protein_version`, which might disagree if the user mixed releases.

2. **Expose a `fasta_version` accessor.** With the version retained we can offer `transcript.protein.fasta_version`, `transcript.fasta_version`, or `Genome.protein_fasta_version("ENSP...")` so downstream tooling can sanity-check FASTA / GTF alignment instead of silently using stale sequences.

3. **Drop the runtime-strip workaround.** Lookups via `lookup_sequence_with_version_fallback` currently do an `.rsplit(".", 1)` on every miss. If both forms are indexed at parse time the helper collapses to a single dict get.

## Proposed shape

`SequenceData` already builds a `{identifier → sequence}` dict from the FASTA. Replace it with:

- `self._fasta_dictionary: dict[str, str]` — the canonical map. Key on the **versioned** ID when the FASTA header carried a version, else the bare ID. (No data loss.)
- `self._stripped_index: dict[str, str]` — maps `ENS<id>` → `ENS<id>.N` when a versioned alias exists. Built once during parse.

Then `lookup_sequence_with_version_fallback(seq_data, id)` becomes:

```python
seq = seq_data._fasta_dictionary.get(id)
if seq is not None:
    return seq
versioned = seq_data._stripped_index.get(id)
if versioned is not None:
    return seq_data._fasta_dictionary[versioned]
return None
```

`SequenceData` also gains a `fasta_version(id)` accessor returning the integer suffix from the FASTA header (or `None` if the FASTA didn't carry one).

## Cost

- One extra dict (~one entry per ENS-prefixed FASTA record, plus a few hundred KB of string objects per FASTA).
- The pickle written under the download cache grows proportionally — invalidate the pickle on schema bump (already tied to `DATABASE_SCHEMA_VERSION` via the pickle filename pattern).
- Constructor signature: `SequenceData` is built internally by `Genome`; no public API break.

## Why defer to its own PR

This is a real change to `SequenceData`'s data model — worth designing deliberately rather than tucking into a bugfix. Good pairing candidate with **#337** (chromosome-FASTA tier) since that work also has to design `Genome.sequence(contig, start, end, strand)` indexing semantics; if both land in the same release the parser + indexer can be reasoned about together.

## Scope guard

Out of scope for this issue:
- Non-ENS IDs (e.g. TAIR `AT1G01010.1` — `.N` is an isoform suffix, not a version). Behavior for those stays as-is.
- Migration of the lookup helper's call sites — those just keep working with the new internal layout.

iskandr:
## Summary

- **`_parse_header_id`** stops stripping ENS `.N` version suffixes — the versioned form becomes the canonical FASTA identifier. Also handles GENCODE pipe-delimited headers properly (leading `ENSP00000493376.2` instead of the whole pipe-packed blob with everything after the first dot lopped off).
- **`SequenceData`** gains two parse-time data structures:
  - `_stripped_index: bare_ens_id -> versioned_id` so bare-ID lookups still resolve against a versioned FASTA (the GENCODE case)
  - `_versions: id -> int` exposed via a new `fasta_version(id)` accessor for downstream sanity-checking of FASTA / GTF alignment
- **`lookup_sequence_with_version_fallback`** resolves both directions (versioned caller -> bare FASTA, bare caller -> versioned FASTA) and drops the per-miss `.rsplit` workaround. Falls back gracefully when handed an old `SequenceData` (e.g. from an unpickled object) that lacks `_stripped_index`.
- Pickle filename bumped from `<fasta>.pickle` to `<fasta>.v2.pickle` (constant `FASTA_PICKLE_SCHEMA_VERSION`) so caches from the v1 bare-key layout get rebuilt instead of silently loaded.
- Version bumped to 2.10.0.

Companion to **openvax/gtfparse#67** (attribute_aliases + cast_version_columns). Together these two PRs close out the GENCODE compatibility regression originally surfaced in #335 — the user's example (`Variant(...).effects()` returning NoncodingTranscript for every transcript) becomes pure data-flow once both are released.

## Test plan

- [x] `./lint.sh` clean
- [x] `./test.sh` — 199 passed (22 new), no regressions; existing `test_versioned_protein_fasta.py` (#335 part-2 tests) still green.
- [x] `tests/test_fasta_versions.py` covers:
  - `_parse_header_id` keeps versioned ENS ID; bare ENS unchanged; splits on space (Ensembl) and pipe (GENCODE); doesn't touch TAIR `AT1G01010.1`
  - `_split_ens_version` extracts the int version, returns None for bare or non-ENS, defensive for unparseable suffix
  - `SequenceData` keys versioned IDs verbatim; `_stripped_index` maps bare ENS → versioned; `fasta_version()` accepts either form
  - bare-key fallback when header carries no version
  - TAIR isoform suffix is keyed as-is and bare lookup returns None
  - conflict resolution: when two versioned forms of the same bare ENS coexist, the higher version wins in `_stripped_index`
  - pickle filename contains `FASTA_PICKLE_SCHEMA_VERSION`; pickle round-trip rebuilds `_stripped_index` from disk
  - `lookup_sequence_with_version_fallback` resolves direct hits, versioned-caller→bare-FASTA, bare-caller→versioned-FASTA, and returns None for unknown/empty inputs
  - TAIR `.1` isoform is NOT stripped by the fallback (would be semantically wrong)
  - legacy `SequenceData` without `_stripped_index` still resolves via literal dict get
  - pickled FASTA dictionary on disk uses versioned keys

### DIFF
--- a/pyensembl/database.py
+++ b/pyensembl/database.py
@@ -631,6 +631,14 @@ class Database(object):
             # 2.7.0+) renames the GENCODE columns onto the Ensembl names at
             # parse time so the rest of pyensembl can stay Ensembl-centric.
             attribute_aliases=GENCODE_BIOTYPE_ALIASES,
+            # Opt out of gtfparse's Int64 cast for *_version columns: when
+            # any row is missing a version (e.g. start_codon rows that
+            # don't carry transcript_version), pandas' nullable Int64
+            # routes through float on the sqlite write path and stores as
+            # "7.0" text, which then breaks our `int(result[0])` parse.
+            # pyensembl handles the string→int conversion itself on the
+            # `*.version` property side.
+            cast_version_columns=False,
             infer_biotype_column=True,
             usecols=usecols,
             features=features,
--- a/pyensembl/fasta.py
+++ b/pyensembl/fasta.py
@@ -29,7 +29,17 @@ logger = logging.getLogger(__name__)
 def _parse_header_id(line):
     """
     Pull the transcript or protein identifier from the header line
-    which starts with '>'
+    which starts with '>'.
+
+    The full versioned form (e.g. ``ENSP00000123456.3``) is returned when
+    the header carries a version. Stripping happens at lookup time via
+    :func:`pyensembl.sequence_data.lookup_sequence_with_version_fallback`,
+    not here, so the FASTA-header version is preserved as the authoritative
+    identity of the sequence.
+
+    Non-ENS IDs (e.g. TAIR ``AT1G01010.1``, where ``.1`` is an isoform
+    suffix rather than a version) are returned verbatim — this function
+    does no surgery on them.
     """
     if type(line) is not bytes:
         raise TypeError(
@@ -47,20 +57,34 @@ def _parse_header_id(line):
     else:
         identifier = line[1:]
 
-    # annoyingly Ensembl83 reformatted the transcript IDs of its
-    # cDNA FASTA to include sequence version numbers
-    # .e.g.
-    # "ENST00000448914.1" instead of "ENST00000448914"
-    # So now we have to parse out the identifier
-
-    # only split name of ENSEMBL naming. In other database, such as TAIR,
-    # the '.1' notation is the isoform not the version.
-    if identifier.startswith(b"ENS"):
-        dot_index = identifier.find(b".")
-        if dot_index >= 0:
-            identifier = identifier[:dot_index]
+    # GENCODE protein/cDNA FASTAs use a pipe-delimited header packing the
+    # versioned protein ID, transcript ID, gene ID, etc., e.g.
+    #   >ENSP00000493376.2|ENST00000641515.2|ENSG00000186092.7|...
+    # The leading field is the canonical identifier for this sequence;
+    # everything after the first '|' is cross-reference metadata.
+    pipe_index = identifier.find(b"|")
+    if pipe_index >= 0:
+        identifier = identifier[:pipe_index]
 
     return identifier.decode("ascii")
+
+
+def _split_ens_version(identifier):
+    """
+    Split an ENS-prefix identifier into ``(bare_id, version_int)``.
+
+    Returns ``(identifier, None)`` for IDs that don't carry a parseable
+    ENS version. Non-ENS IDs (e.g. TAIR ``AT1G01010.1``) are always
+    returned as-is with version ``None`` — the ``.N`` in those is an
+    isoform suffix, not a version.
+    """
+    if not identifier or not identifier.startswith("ENS") or "." not in identifier:
+        return identifier, None
+    bare, _, suffix = identifier.rpartition(".")
+    try:
+        return bare, int(suffix)
+    except ValueError:
+        return identifier, None
 
 
 class FastaParser(object):
--- a/pyensembl/protein.py
+++ b/pyensembl/protein.py
@@ -19,9 +19,14 @@ class Protein(Serializable):
     Accessed via :attr:`Transcript.protein`.
     """
 
-    def __init__(self, protein_id, protein_version=None):
+    def __init__(self, protein_id, protein_version=None, genome=None):
         self.protein_id = protein_id
         self.protein_version = protein_version
+        # Optional genome reference so ``fasta_version`` can consult the
+        # protein FASTA's :class:`SequenceData`. Defaults to ``None`` for
+        # back-compat (callers constructing ``Protein`` outside of
+        # :class:`Transcript` won't have a meaningful genome).
+        self.genome = genome
 
     @property
     def id(self):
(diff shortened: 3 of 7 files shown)


iskandr (commented):
Two requested changes before landing — bundling rather than spinning a follow-up:

## 1. Drop the pickle filename bump

The `FASTA_PICKLE_SCHEMA_VERSION = "v2"` rename is overcautious. **v1 pickles are forward-compatible with this code.**

Walking through what happens when v2.10.0 loads an old (bare-keyed) v1 pickle:

- `_add_to_fasta_dictionary` iterates the loaded dict.
- For each `(identifier, sequence)`: identifier is bare (`ENSP00000123456`), so `_split_ens_version` returns `(id, None)`.
- The `if version is not None:` guard skips both `_stripped_index` and `_versions` writes.
- Result: bare-keyed `_fasta_dictionary`, empty `_stripped_index`, empty `_versions`.

That state still resolves every realistic lookup correctly:

| Caller passes | Dict has (v1 cache) | Result |
|---|---|---|
| bare `ENSP00000123456` | bare key | direct hit ✓ |
| versioned `ENSP00000123456.3` (GENCODE GTF) | bare key | step-2 strip-and-retry hits ✓ |
| bare against versioned-FASTA (the new GENCODE→bare path) | n/a — only happens after a fresh parse, which is fine |

The only capability lost on a stale v1 cache is `fasta_version(id)` returning a real int — but that information was never captured by the v1 parser, so returning `None` is honest, not wrong.

**Suggested patch:** drop `FASTA_PICKLE_SCHEMA_VERSION` and revert the filename change in `SequenceData.__init__`:

```python
self.fasta_dictionary_filenames = [
    filename + ".pickle" for filename in self.fasta_filenames
]
```

Plus a regression test that constructs an old-style bare-keyed pickle on disk and asserts `SequenceData` loads it without erroring, and that `lookup_sequence_with_version_fallback` still resolves both bare and versioned callers against it.

## 2. Surface `fasta_version` on the user-facing view objects

`SequenceData.fasta_version(id)` is exposed but nothing in the public API plumbs it up. Easy add now, harder to retrofit later if downstream tools start caching the wrong protein version.

**`Transcript.fasta_version`** — straightforward, `Transcript` already has `self.genome`:

```python
@property
def fasta_version(self):
    """Version embedded in this transcript's cDNA FASTA header, or None."""
    if self.genome.transcript_sequences is None:
        return None
    return self.genome.transcript_sequences.fasta_version(self.transcript_id)
```

**`Protein.fasta_version`** — needs a genome reference on the `Protein` view, which it doesn't carry today. Smallest change: thread `genome=` through `Protein.__init__` and the `Transcript.protein` constructor:

```python
# protein.py
class Protein(Serializable):
    def __init__(self, protein_id, protein_version=None, genome=None):
        self.protein_id = protein_id
        self.protein_version = protein_version
        self.genome = genome

    @property
    def fasta_version(self):
        if self.genome is None or self.genome.protein_sequences is None:
            return None
        return self.genome.protein_sequences.fasta_version(self.protein_id)

# transcript.py — Transcript.protein
return Protein(
    protein_id=self.protein_id,
    protein_version=protein_version,
    genome=self.genome,
)
```

Tests to add:
- `transcript.fasta_version` returns the int when the cDNA FASTA carried a version; None when it didn't.
- `transcript.protein.fasta_version` returns the int when the protein FASTA carried a version.
- A consistency-check pattern: `transcript.protein.protein_version` (from GTF) vs `transcript.protein.fasta_version` (from FASTA) — useful when a user mixes a fresh GTF with a stale FASTA.

This matches the rationale I gave in #351 for keeping the version in the first place: the FASTA-header version is the authoritative source-of-truth for sequence identity, and downstream consumers want to be able to *see* it.

---

Both changes are small. Reverting the schema bump is ~4 lines + a test. Adding `fasta_version` on the views is ~15 lines + tests. Together they make the PR fully close out the original issue rather than leaving the user-facing accessor as a follow-up.

Issue #367, opened 2026-09-14
Locus/Exon comparisons raise AttributeError for non-Locus operands


iskandr:
`Locus.__eq__`, `__lt__`, and `__gt__` (`pyensembl/locus.py:72,88,99`) and `Exon.__eq__` (`pyensembl/exon.py:82`) build their TypeError message with `other.__class.__name__`. Inside a class body that name gets mangled to `other._Locus__class`, so comparing with a non-Locus raises AttributeError before the TypeError is ever built:

```
>>> from pyensembl import Locus
>>> Locus("1", 10, 20, "+") == None
AttributeError: 'NoneType' object has no attribute '_Locus__class'
```

There's a bigger question than the typo. With `__eq__` raising for foreign types, `locus == None`, `locus in [None, ...]`, and mixed-type dict or set lookups all throw. `Gene.__eq__`, `Transcript.__eq__`, and `Protein.__eq__` return `False` in that case, so the classes don't agree with each other. The usual contract is to return `NotImplemented` for `__eq__`/`__lt__`/`__gt__` and let Python fall back.

The typo fix (`__class__`) is safe. Switching `==` from raising to returning False changes behavior that downstream code (varcode compares Exons heavily) might rely on, so I'm asking before changing it.

PR #381, opened 2026-09-15
Fix name-mangled other.__class typo crashing Locus/Exon comparisons

agu2347:
Fixes the crash from #367.

## Root cause

`Locus.__eq__`, `__lt__`, `__gt__` (`pyensembl/locus.py`) and `Exon.__eq__` (`pyensembl/exon.py`) build their `TypeError` message with `other.__class.__name__` instead of `other.__class__.__name__`. Written inside a class body, `other.__class` (two leading underscores, zero trailing) gets name-mangled to `other._Locus__class` / `other._Exon__class` -- which essentially no object has. So comparing a `Locus` or `Exon` with anything that isn't one raises `AttributeError` before the intended `TypeError` is ever built:

```pycon
>>> Locus("1", 10, 20, "+") == None
AttributeError: 'NoneType' object has no attribute '_Locus__class'
```

## Fix

Added the missing trailing `__` in all four spots.

As @iskandr noted in #367, there's a separate, broader question about whether `__eq__` should return `NotImplemented` instead of raising for foreign types (so `locus in [None, ...]`, mixed-type set/dict lookups, etc. work), since `Gene`/`Transcript`/`Protein` already return `False` in that case rather than raising. That's a real behavior-change question `varcode` (a heavy consumer of `Exon` equality) may care about, and the issue explicitly asked for maintainer input before making that call -- so this PR only fixes the crash and leaves that design decision alone.

## Testing

Added to `tests/test_locus.py`:
- `test_locus_comparison_with_non_locus_raises_type_error_not_attribute_error`: `Locus == None`, `Locus < None`, `Locus > None`, `Locus == 42` now raise `TypeError` (previously `AttributeError`).
- `test_exon_comparison_with_non_exon_raises_type_error_not_attribute_error`: same for `Exon`.

Verified via `git stash` that both new tests fail with the exact reported `AttributeError` against the unpatched code, and pass after the fix. Ran `tests/test_locus.py` (10/10 pass) and `tests/test_exon_id.py` in full -- the 3 pre-existing failures there are unrelated (they need `pyensembl install` to fetch real Ensembl release data) and fail identically on unpatched `main`.

---

_Disclosure: this PR was prepared with AI assistance (Claude), used under my direction as the account owner. I reviewed the root-cause analysis, diff, and tests before opening it._


iskandr:
Reproduced and merged as #391 (same diagnosis, same fix, same tests) from a same-repo branch so CI could actually run — this PR is from a fork and GitHub was holding the workflow pending approval, so it never got a CI signal. Closing this one in favor of #391, credited there. Thanks for the fix.

PR #387, opened 2026-09-15
CLI: one-line errors instead of tracebacks (2.10.14)

iskandr:
Common CLI mistakes printed a full Python traceback. They now print one line and exit 1.

Before:
```
$ pyensembl install --release 999
Traceback (most recent call last):
  ... 20 lines ...
ValueError: No genome for homo_sapiens in Ensembl release 999
```
After:
```
$ pyensembl install --release 999
pyensembl: error: No genome for homo_sapiens in Ensembl release 999
```

Covered: unknown release, unknown species, `--gtf` without `--reference-name` or `--annotation-name`, `--release` combined with `--gtf`, a missing local GTF/FASTA (`MissingFile(/path)` becomes `file not found: /path`), and `delete-index-files` for a genome that isn't installed. `--transcript-fasta` without `--gtf` used to raise `TypeError` from `os.path.join(prefix, None)`; it now works.

Help and usage:
- `usage=__doc__` made every argparse error (`pyensembl frobnicate`, `--release abc`) print the whole 25-line docstring before the error. The docstring is now the `--help` description, so errors show the normal short usage.
- The install-from-source example lost its line continuations (`install  --reference-name "GRCh38"  --gtf ...` on one line) and omitted `--annotation-name`, which is required. Both fixed.

Bumps version to 2.10.14 (#386 already takes 2.10.13).

Tests: `./lint.sh`; `python -m pytest tests/test_shell.py -n 0 -q` (14 passed). The 6 new parametrized cases fail on main.

https://claude.ai/code/session_01DctgBZ5rJiuNzpWXqrNguT


iskandr:
**Verdict:** merge after fixes. The user-facing change is right and all six new cases really do fail on `main`, but the version bump has gone stale and the `except ValueError` is wide enough to hide corrupt-data failures.

## Verified

Ran everything the PR body claims, against the branch at `fc2f2cb`:

- `./lint.sh` — passes.
- `python -m pytest tests/test_shell.py -n 0 -q` — 14 passed.
- The 6 new parametrized cases fail on `main`: copied `tests/test_shell.py` onto a worktree at `8eeb302` and got `6 failed, 8 passed`. Holds.
- Each error path by hand, via the installed `pyensembl` console script (so `prog` is the real one, not `-c`):

| command | exit | stderr |
|---|---|---|
| `install --release 999` | 1 | `pyensembl: error: No genome for homo_sapiens in Ensembl release 999` |
| `install --species martian` | 1 | `pyensembl: error: Species not found: martian, ...` |
| `install --gtf x.gtf` | 1 | `pyensembl: error: --reference-name is required with ...` |
| `install --reference-name GRCh38 --gtf x.gtf` | 1 | `pyensembl: error: --annotation-name is required with ...` |
| `install --release 81 --reference-name GRCh38 --annotation-name test --gtf x.gtf` | 1 | `pyensembl: error: --release cannot be combined with ...` |
| `install --reference-name GRCh38 --annotation-name test --transcript-fasta /nonexistent/t.fa` | 1 | `pyensembl: error: file not found: /nonexistent/t.fa` |
| `delete-index-files --release 75` (not installed) | 1 | `pyensembl: error: Missing genome data file from https://... Run pyensembl install ...` |
| `frobnicate` | 2 | short usage + `invalid choice` |
| `install --release abc` | 2 | short usage + `invalid int value: 'abc'` |

- `--help` renders the docstring as the description: `%(prog)s` is substituted, the `r"""` keeps the `\` continuations, and the install-from-source example now shows `--annotation-name`. Argument errors print the 12-line usage instead of the 25-line docstring. All as claimed.
- `--transcript-fasta` without `--gtf` no longer raises `TypeError`; a `Genome` with `gtf_path_or_url=None` is built and indexing the FASTA works.
- **Partial install:** ran the same matrix on this branch and on `main` — identical results, so no regression here. A clean install, a re-install over a complete one, and an install over a 0-byte `.db` all exit 0 and (re)build the index. An install over a *truncated* or garbage `.db` fails with `sqlite3.DatabaseError: database disk image is malformed` on both branches, and `--overwrite` does not help on either. Pre-existing; see nit 4.
- **#390:** this PR does not touch that path. `collect_selected_genomes` / `all_combinations_of_ensembl_genomes` are unchanged apart from the new error types, and `pyensembl delete-all-files` with no `--release` still logs `Running 'delete-all-files' for EnsemblRelease(release=115, species='homo_sapiens')` and exits 0. The only #387 effect is the one #390 already records: the `delete-index-files`-on-an-uninstalled-genome traceback becomes a one-liner, with the same backwards wording.

## Problems

**1. The version bump is stale: `main` is already 2.10.14.** `pyensembl/version.py:1`

#386 took 2.10.13, but #391 (`8eeb302`, "Fix name-mangled other.__class typo") merged afterwards and took 2.10.14:

```
$ git show origin/main:pyensembl/version.py | head -1
__version__ = "2.10.14"
```

Merging as-is leaves `main`'s version unchanged, which breaks AGENTS.md rule 2 ("every PR bumps the version") and means `./deploy.sh` would try to tag `v2.10.14` twice. 2.10.15 is free — it is the only open PR.

**2. `except ValueError` also swallows genuine data-integrity failures.** `pyensembl/shell.py:400`

`ValueError` is not a "bad command line" marker in this codebase — it is the generic error type. `database.py:190` and `:195` raise it when a GTF's rows can't form a primary key, `database.py:466` when a query returns nothing, `genome.py:822` and `:1011` for `"Result is not the expected length"`, plus `normalization.py` and `locus.py`. Several of those sit directly on the `install` path, and they now print as if the user mistyped something:

```
$ pyensembl install --reference-name R --annotation-name test --gtf dup_gene_id.gtf
pyensembl: error: Column 'gene_id' can't be primary key of table 'gene' because it contains repeated values
$ echo $?
1
```

(`dup_gene_id.gtf` is a 4-line GTF whose two `gene` rows share a `gene_id`.) No traceback, no mention of the GTF path, no mention of which stage failed, and no `--debug` to get the frames back.

It also makes the CLI inconsistent about which failures are readable, since the split is by exception type rather than by whose fault it is. On the same branch, with the same kind of unusable input:

- empty GTF -> full traceback ending `polars.exceptions.NoDataError: empty CSV`
- GTF with the wrong column count -> full traceback ending `gtfparse.parsing_error.ParsingError: Wrong number of columns`
- `--gtf` pointing at a directory -> full traceback ending `IsADirectoryError`
- truncated `.db` -> full traceback ending `sqlite3.DatabaseError`
- duplicate `gene_id` -> one line

**3. Exit code 1 for what argparse (and mhctools) call a 2.** `pyensembl/shell.py:399-401`

The three checks in `collect_selected_genomes` are pure argument validation — they are reached before any I/O and they name flags. They now exit 1, while argparse's own checks in the same program exit 2, and all of them print the identical `pyensembl: error: ` prefix:

```
$ pyensembl install --gtf x.gtf ; echo $?
pyensembl: error: --reference-name is required with --gtf, --transcript-fasta or --protein-fasta
1
$ pyensembl frobnicate ; echo $?
... usage ...
pyensembl: error: argument action: invalid choice: 'frobnicate' ...
2
```

pyensembl had no prior convention here (no test asserted a CLI failure exit code before this PR), so nothing in-repo is broken. But `mhctools` `master` already routes this class of error through `parser.error()` — exit 2 — in six CLI modules, and `parser.exit(1, ...)` reproduces `parser.error()`'s exact message format while changing only its status. Suggest `parser.error()` for the flag-combination checks and for unknown release/species, keeping 1 for runtime failures like a missing file. I have not changed this: it is a deliberate choice in the PR and its tests assert 1.

## Nits

1. `tests/test_shell.py:190` asserts `result.stderr...startswith("-c: error: ")`. `-c` is an artifact of `_run_cli` launching `python -c`; the string a user sees is `pyensembl: error: `. A prefix-agnostic check (or invoking the console script) would not pin the harness into the assertion.
2. `--gtf ""` is falsy, so `collect_selected_genomes` silently ignores `--reference-name` / `--annotation-name` and falls through to the Ensembl path — `pyensembl install --reference-name R --annotation-name test --gtf ""` downloads and indexes human release 115 (3.2 GB here) and exits 0. Pre-existing, but it is the same category of "common CLI mistake" this PR is about.
3. `MissingRemoteFile` is not caught alongside `MissingLocalFile`. That is correct today only because `local_path_or_install_error` converts it to a `ValueError`; it is worth a comment so the asymmetry does not read as an oversight.
4. An install over a truncated or corrupt `.db` fails with a raw `sqlite3.DatabaseError` and `--overwrite` does not recover it (identical on `main`). Worth its own issue rather than this PR.
5. `run_action`'s trailing `raise ValueError("Invalid action: %s" % args.action)` is unreachable — argparse `choices` already rejects anything else.

iskandr:
Pushed 6f49d93, which addresses items 1 and 2 from the review. (The reviewer's own push was blocked by a permission prompt; I pushed it after checking the diff.)

- **Item 1, stale version bump.** `main` is already at 2.10.14: #386 took 2.10.13 and #391 took 2.10.14. This branch now bumps to 2.10.15, so merging changes the version and doesn't re-tag an existing one.
- **Item 2, internal errors hidden.** `except ValueError` was catching pyensembl's generic error type, so a malformed GTF (for example duplicate `gene_id` values in `gene` rows) printed a single line with no file, no stage and no traceback. Now there are two narrow types, `CommandLineError` and `MissingGenomeDataFile`, both subclassing `ValueError` so library callers and existing `raises(ValueError)` tests are unaffected. `ValueError` is converted to `CommandLineError` only while genomes are being selected, which is pure argument interpretation. Indexing and download failures keep their tracebacks, and there's a regression test for that.

Every error message from the original PR is unchanged, and the six new parametrized cases still fail on `main`, so the fix they cover is still real.

Left for your judgment:
- **Exit codes.** Flag-combination mistakes exit 1 while argparse errors exit 2, both printing the same `pyensembl: error:` prefix. mhctools uses `parser.error()` (exit 2) for this class of mistake. I didn't change it, since the PR's tests assert 1 and it's a deliberate choice.
- **`--gtf ""`** silently installs the 3.2 GB default human genome.
- **A truncated or corrupt `.db`** fails with `sqlite3.DatabaseError` and `--overwrite` doesn't recover it. Pre-existing on `main`, not introduced here.

